PR #22, opened 2017-11-24
Fix stu-home.desktop permissions

kylelaker:
The copy module in Ansible does not preserve permissions by default, so
when this file ends up on the user's desktop, the first time they
attempt to execute the file, they get a message telling them that the
file is not executable (trusted). Adjust the permissions on the file to
0755 in order to prevent that message.


kylelaker:
Requesting copy module preserve permissions: ansible/ansible-modules-core#1124

Change to use existing permissions unless a different mode is specified: ansible/ansible@5502da3 which will end up in 2.3; however, Mint currently has 2.1

PR #24, opened 2017-11-24
Fix jgrasp download URL and hash

kylelaker:


kylelaker:
I guess jgrasp doesn't keep the old versions around. 204_01 just 404s

PR #34, opened 2017-12-01
Create OEM install script and playbook

### LINKED ISSUE #11: Create OEM playbook for pre-distribution tasks

### LINKED ISSUE #27: Create welcome/what's next doc on desktop
- [x] Write documentation
- [x] Write .desktop file
- [ ] Create OEM task for .desktop file

kylelaker:
This OEM build script in scripts/ ensures that necessary packages are
installed through apt before attempting to run Ansible. After that, it
runs the playbook for the OEM role installing and removing packages as
necessary and preparing some .desktop files for users to click when they
first sign in. Once the script completes, the VM should be in a state
that is nearly ready to ship to users.

Resolves #11 
Resolves #27


ripleymj:
Despite the fact I shouldn't look at these things late at night, a few thoughts:
- apt install ansible would be nice to have "-y"
- apt update in the oem playbook seems unnecessary
- desktop shortcuts may be nice to have with the JMU tux logo, or something more exciting
- removing "-K" from ansible-pull and prefixing with gksudo in the wrapper may provide a better experience
- when we purge the old Mint themes, we're losing the desktop wallpaper. This seems like an upstream bug.
- I'm still undecided if we need the real VBox tools installed at OEM time

This also leaves open for local.yml, the inclusion of libnotify-bin and creating the Menu shortcuts.

kylelaker:
> apt install ansible would be nice to have "-y"

Got it. Will push shortly.

> apt update in the oem playbook seems unnecessary

Do we want to do it first in the `oem-build` script then?

> desktop shortcuts may be nice to have with the JMU tux logo, or something more exciting
 
I'll look into what file formats/sizes would be necessary

> removing "-K" from ansible-pull and prefixing with gksudo in the wrapper may provide a better experience

Do we also want to just consider making the error message we show when the playbook fails a GTK dialog? If so, then oem-build needs to become a python script. For now, though, I'll add `gksudo` and we can decide more later.

> when we purge the old Mint themes, we're losing the desktop wallpaper. This seems like an upstream bug

gsettings set org.cinnamon.desktop.background picture-uri  "file:///PATH" might be good enough. Maybe we could consider using a JMU background like the CS Dept labs.

> I'm still undecided if we need the real VBox tools installed at OEM time

Not until 5.2 tools are fixed. We don't want people enabling 3D acceleration and breaking boot. Maybe in a future semester.

> We will need to decide when to switch it to use a git branch for ansible-pull

Is this for the OEM scripts as a whole or the `uug-ansible-wrapper`?

kylelaker:
Desktop background issue appears to be that `/usr/share/backgrounds/linuxmint` gets deleted. `/usr/share/backgrounds/linuxmint/default_background.jpg` is the default. None of `mint-background-{sarah,serena,sonya} provide that directory or a file `default_background.jpg`

PR #44, opened 2017-12-03
Purge checkout, clean up formatting

ripleymj:
Purge checkout after running to avoid problems with stale data. Also, move URL out of line to clean up formatting.

### DIFF
--- a/roles/oem/files/uug-ansible-wrapper
+++ b/roles/oem/files/uug-ansible-wrapper
@@ -1,12 +1,14 @@
 #!/usr/bin/env bash
 # bash is used because read -p and echo -e are undefined in POSIX sh
 
+REPO_URL="https://github.com/jmunixusers/cs-vm-build"
+
 if [ -z "$1" ]; then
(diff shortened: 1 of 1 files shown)


roles/oem/files/uug-ansible-wrapper
```diff
@@ -1,12 +1,14 @@
...
 
+REPO_URL="https://github.com/jmunixusers/cs-vm-build"
+
 if [ -z "$1" ]; then
     read -p "Re-run this as root. Press ENTER to exit" junk
     exit 1;
 fi
 
-if ansible-pull -U https://github.com/jmunixusers/cs-vm-build -i hosts -t "$1"; then
+if ansible-pull --url $REPO_URL --purge -i hosts -t "$1"; then
```
kylelaker: It may be preferable to wrap `$REPO_URL` in quotes just to prevent potentially future issues with spaces or hyphens.

kylelaker:
Feel free to merge this, my comment isn't worth blocking the PR over. Just thoughts in case we end up using more complex URLs in the future or a shell acts dumb.

PR #48, opened 2017-12-03
Menu target shortcuts

### LINKED ISSUE #41: Add course targets to program menu, like OEM task for desktop

ripleymj:
Sorry, this ended up being a huge commit. Potentially resolves #41. No pressure to merge if it looks horrible.


ripleymj:
I was able to run it in both common and oem after working out commit 2 in the series. I'm just concerned about moving that much functionality around in one commit, and trying to do it early in the morning.

kylelaker:
Reviewing this inspired #49; however, I am going to consider that as a separate issue. This seems like a very clean way of going about doing this. The files get installed to the proper locations, it keeps `oem` and `common` distinct, and prevents  repetition between those two roles.

PR #53, opened 2017-12-03
Test using Zenity for interacting with user

### LINKED ISSUE #50: uug-ansible-wrapping not waiting
The end of the ansible wrapper script, where we wait for keyboard input, does not seem to fire when run from a desktop icon. Perhaps gksudo isn't running bash interactively, which would disable prompts.

kylelaker:
Fixes #50


kylelaker:
When testing, I think the `common` role may be overwriting the `uug-ansible-wrapper` script, so running a second .desktop file won't work until we either a) add support for other branches in the script or b) merge this into master

kylelaker:
Yep. Definitely thought I remembered that block was doing something else. I'll clean that up and submit a new patch.

PR #71, opened 2017-12-08
Add filezilla

### LINKED ISSUE #69: Include Filezilla, possibly drop Home on Stu
Per @ChrisMayfield, sftp via vfs is unnecessarily confusing for early students (101/149/159) and remote file access is better accomplished with FileZilla. Adding FileZilla is easy enough, bonus points if we pre-populate a connect entry in its configuration. Need to determine if we offend @lam2mo by dropping the sftp-vfs across the board in favor of sftp.

kylelaker:
Resolves #69

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ apt will require more extensive modifications.
 * JDK/JRE 8
 * Wireless printing
 * Shortcuts to re-run tasks
+* Filezilla
 
 ### CS101
 * Mint packages (basic-prog-pkgs role)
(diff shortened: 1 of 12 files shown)


kylelaker:
As it stands now, `stu-home.desktop` still exists in the repo, there is just no task to install it anywhere.

ripleymj:
Tracking stu-home in #72

kylelaker:
There was no good reason other than not thinking that through. Fixed in 4636211

PR #80, opened 2017-12-13
Remove pygame and aptana

Advill:
Removed Aptana and Python-pygame from automatic install lists.


kylelaker:
Thanks, Mike! Great job!

PR #88, opened 2017-12-18
Filezilla: Hide dotfiles by default

### LINKED ISSUE #87: Filezilla does not hide dot files by default
When selecting the filter, it seems to create this as .filezilla/filters.xml: https://gist.github.com/ripleymj/1554cf8806c0779e09873c0f45c7a82b

kylelaker:
Fixes #87


kylelaker:
Justification for force=no on the copy task is because if a user goes in to disable those filters and then runs the role again, their changes will be reverted which is bad. Additionally, any new filters they add would be deleted. The unfortunate thing would be if the user already had a `.filezilla/filters.xml` but it was either empty or didn't have these filters enabled by default (such as if the user installed Filezilla before running any roles). I don't think that is worth the risk of overwriting a config.

PR #90, opened 2017-12-19
Install Zenity as a dependency

### LINKED ISSUE #89: Install Zenity as a dependency
This should be installed in Mint by default; however, since it will be critical to the errors and warnings in the Ansible wrapper script, we should try our best to ensure it is installed. It should definitely be added to the OEM role; however, I am uncertain on the value of adding it to common.

kylelaker:
Resolves #89

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -46,6 +46,7 @@
     - libnotify-bin
     - openjdk-8-jdk
     - openjdk-8-source
+    - zenity
 - name: Touch VM profile
   copy:
     src: csvmprofile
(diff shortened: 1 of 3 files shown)


ripleymj:
Pushed a small tweak for your consideration

ripleymj:
Also, I swear I didn't do this so that we'd need an unbiased review from a 3rd party.

kylelaker:
Rebased this atop master to resolve merge conflicts

kylelaker:
@ripleymj are we still waiting for a review from @Advill before merge?

PR #62, opened 2017-12-05
Use data from /etc/os-release to determine what branch to pull

kylelaker:
The VERSION_CODENAME is how we will be tagging branches in the future.
Source /etc/os-release to get the VERSION_CODENAME variable and pass
that to ansible-pull's --checkout option. Report an error if
/etc/os-release does not exist.

This must not be merged until we create a branch for the Sylvia release.

Issue #84, opened 2017-12-18
Install vim and unzip by default


kylelaker:
The following packages need to be added to the "install dependencies" task of the common role:

* vim
* unzip

ripleymj:
`vim` should also be removed from the adv-prog-pkgs role if moved to common.

Issue #65, opened 2017-12-07
USB hardware level


ripleymj:
Per @ChrisMayfield, we should be intentional about requiring the Extension Pack when packaging the VM, as most people won't need it. CS354 may require a better USB controller for moving Kinect data into the VM.

- [x] Remove USB 2.0 from the next image build
- [x] Add a warning/failure if EHCI/xHCI controllers exist during OEM script
- [x] Find a way to warn in the CS354 target of a missing EHCI/xHCI controller
- [x] Create a docs folder, and publish a How to Add Extensions and Upgrade USB markdown file

kylelaker:
Do we want the EHCI/xHCI warning to be a libnotify notification or did you have something else in mind?

ripleymj:
In theory, I like that we can do libnotify and move on, but I feel this may be worthy of the subtlety that zenity offers.

kylelaker:
As far as the warning, I am not sure how to best go about it, so I'm hoping you have something in mind. The best I have though of is:

`roles/robot-pkgs/tasks/main.yml`
```
- name: Check for missing USB 2/3 controllers
  shell: lspci | grep -i -e EHCI -e xHCI
  register: lspci_output
  #failed_when: lspci_output.rc != 0
  changed_when: False
- name: Alert missing USB 2/3 controllers
  notify:
      - USB controllers warning
  when: lspci_output.rc != 0
```
`roles/robot-pkgs/handlers/main.yml`
```
- name: USB controllers warning
  command: >-
    zenity --warn --text "USB Controllers missing"
    "In order to effectively connect robots to the VM"
    "ECHI/xHCI controllers need to be installed from"
    "VirtualBox. See <a href='URL'>these instructions</a> for"
    "more information on how to install them"
  become: yes
  become_user: "{{ item.user }}"
  with_items: "{{ real_users }}"
```
I feel like there is probably a *much* cleaner way of going about it. It feels dirty that checking the condition and notifying the handler are two separate tasks.

ripleymj:
That's about it, and I think the two steps are just an Ansible "feature". The unfortunate thing is that HREF in zenity will likely open as root.

kylelaker:
We'll have to test that. Hopefully if we become_user then it will work out. Though, I am also concerned about how that will play out if there's a user in real_users other than the currently logged-in user and if more than one window comes up.

ripleymj:
I misread become_user, you're probably right that this is ok. Multiple could get interesting though.

[...]

PR #94, opened 2017-12-24
Rewrite wrapper script

### LINKED ISSUE #89: Install Zenity as a dependency

### LINKED ISSUE #92: Enable logging
Before we ship to users we should consider some method of storing logs of runs. None of them are perfect. A few options include:
 * using `tee` in the wrapper script
* setting `log_path` in `ansible.cfg` [[1]]
* setting an `ANSIBLE_LOG_PATH` environment variable in the wrapper

|  |  Pros | Cons |
|-| -----|------|
|`tee`| We can create files, directory ahead of time; we can use shell variables in the file path easily, can log based on tags passed  | It's another command to have to work with and worry  about mucking with exit codes; only works when the wrapper is used  |
|`log_path`| Works regardless of whether or not wrapper is used | More difficult to configure name dynamically (date, time, etc) |
|`ANSIBLE_LOG_PATH`| Benefits of `tee` plus we can allow users to set the environment variable themselves and not override if they've set it  | Only works when the wrapper is used |

We could use a combination of `log_path` and `ANSIBLE_LOG_PATH`. Since Ansible checks the environment first (see [2] after the list of links), we could set a log file in the wrapper `/opt/vm/logs/$TAG/$(date '+%Y%m%d-%H%M%S').log`. In the event that doesn't work, we can have `log_path` set to `/opt/vm/logs/last_run.log`, or a more-likely-to-exist directory `/var/logs/vm_ansible.log`. I really don't want to use `tee` unless there is a compelling reason to do so.

I would like to ship this in the spring VM so that we can get logs from users if (*when*) things go wrong; however, I will let someone else target this to a milestone. It doesn't seem like it'd be awful to implement, but deciding the right way to do it is the trickier part. A nice thing is that all of them are easily changeable since they'd all be set in files we control.

[1]: http://ansible-manual.readthedocs.io/en/latest/intro_configuration.html#log-path
[2]: http://ansible-manual.readthedocs.io/en/latest/intro_configuration.html

### LINKED ISSUE #93: Convert install script to Python
Let's move discussion about this to its own issue, instead of burying it in a PR or comments on a gist: https://gist.github.com/kylelaker/8e8bc9f97514bcda6bfa813d7e7a6f72

kylelaker:
Rewrite the Ansible wrapper script in Python to allow the user to select
which courses they need to have their machine configured for. Enable
logging by specifying a path in ansible.cfg. Remove previous wrappers
and installed desktop shortcuts in favor of the new wrapper

Resolves #93 
Resolves #92 

Closes #62


kylelaker:
@ripleymj Forced pushed a chance to fix that

ripleymj:
I think this works well enough to merge, but I'm going to hold off until later today when I can take one more look at the whole thing.

kylelaker:
@ripleymj  I pushed a change to rebase this on top of master to avoid merge conflicts

kylelaker:
Forgot to comment but the latest version has refactored the dialog creation out to a separate method. It doesn't clean stuff up too much, but hopefully it makes the dialogs slightly more uniform

ripleymj:
I've approved the changes. When you're ready, please merge, create a branch, and tag beta1.

kylelaker:
This can be merged once we have something for #84 and #65.

ripleymj:
Need to rethink the effect on #90 here

kylelaker:
Good point. Due to #97, #89 and #90 are no longer closed by this. I will remove them from the initial message to prevent those from being closed upon merge.

kylelaker:
I agree on the Cancel vs Quit

kylelaker:
Two more commits. One to fix an issue that may come up soon and another to make things cleaner should strings need to be updated in the future.

kylelaker:
Hopefully those are two of the last commits that need to be pushed (and one of the last reviews you'll need to do). When we get ready to merge, I'll rebase this whole thing on master and squash it all into one big commit. Any other changes before then I'll keep pushing as individual commits because I think they more clearly highlight what changes each time I push and make you have to re-review.

kylelaker:
Once all other remaining PRs are merged and we get a resolution to #84 I think we can go ahead and merge this monster as well

kylelaker:
This has been rebased on top of master and all the commits have been squashed.

PR #95, opened 2017-12-25
OEM: Create /usr/local/share/applications

### LINKED ISSUE #70: Add local applications folder to OEM role
If Cinnamon isn't detecting new apps because the /usr/local/applications folder isn't there on first boot, let's try adding it to the OEM role, and then populate it in later roles.

ripleymj:
Create this directory when running the OEM steps in the hope that when
the user runs roles for courses later and the .desktop files get
installed that Cinnamon will detect them automatically (without a
reboot).

Closes #70


kylelaker:
This will result in a merge conflict with #94. Which would you prefer be merged first?

ripleymj:
Let's get this one in first and be done with it.

PR #97, opened 2017-12-26
Alert in cs354 if USB2.0/3.0 controller not present

kylelaker:
Completes one more step in #65

### DIFF
--- a/roles/robot-pkgs/handlers/main.yml
+++ b/roles/robot-pkgs/handlers/main.yml
@@ -1,2 +1,10 @@
 ---
 # handlers file for robot-pkgs
+
+- name: USB controllers warning
+  command: >-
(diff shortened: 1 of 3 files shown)


kylelaker:
This should be tested for the following:

* [ ] More than one user
* [ ] Does it open the URL as root
* [ ] Does the text actually display properly

kylelaker:
With more than one user, you get an error message, but adding "ignore_errors: yes" to the handler prevents that from marking the whole run as failed. It doesn't seem like it causes multiple dialogs to come up

kylelaker:
Depends on #90

ripleymj:
Seems good for now. If we find more needs, for fall, we can consider a step in common to find the UID of the wrapper or cinnamon-session, or pass a preferred UID from the wrapper as an extra variable in the ansible-pull run.

PR #111, opened 2018-02-19
Use JMU CS's mirror for Ubuntu and Linux Mint

kylelaker:


kylelaker:
Should work, but plz to test first

kylelaker:
We may be able to use the HTTPS URLs. We should pay attention to this with #117, but the SP18a image had `apt-transport-https` installed. It doesn't seem like it was pulled in by any dependencies, so it may be a part of the base Linux Mint image (but it does not appear to be a default package in Ubuntu Bionic).

Off GitHub, we had previously discussed not using HTTPS URLs due to the requirement of the extra `apt-transport-https` package.

Once Bionic hits officially (or at least beta) and Mirror is live, I will perform some tests with HTTPS and HTTP mirrors with and without the package installed (and listing all mirrors in `/etc/apt/sources.list`) to see if `apt` can figure out what it needs to do.

My preference would then be to list the mirrors in the following order:

1. JMU CS HTTPS
1. JMU CS HTTP
1. us.archive.ubuntu.com

If HTTPS causes issues, then I'd prefer to drop it.

ripleymj:
Probably time to rethink this one. I think HTTPS works well enough to leave out HTTP, but I also like the idea of having a fallback to upstream. I don't think our current loop in this commit is a good way to handle removing many mirrors, and then readding them with multiple source URLs. Perhaps this would be better delivered as a template file that we overwrite completely on each run.

I also think we need to add something here to account for a shifting upstream ubuntu/mint version, and not hammer the current sylvia/xenial values in from our master branch.

ripleymj:
Seems Ubuntu populates apt/sources.list, where Mint is using apt/sources.list.d/, so we may need to make a call on what distros we intend to support.

kylelaker:
I think that's an important decision to make and perhaps worthy of a full issue unto itself. I am unsure of your specific feeling; however, I think a reasonable way forward would be:
- master targets the latest release of Mint
- Branches target the release of Mint with the same name codename as the branch

Each branch ideally would be able to run on the Ubuntu LTS release that serves a Mint version's base with maybe some minor exceptions. If by some miracle the rest works on Debian or LMDE that's awesome, but not intentional. I think if we receive patches that allow the project to work better on Ubuntu or Debian that's awesome, but that we should not sacrifice functionality on Mint to achieve that goal.


If we want to be compatible with both Ubuntu and Mint, we can read `lsb_release -is` or the fields in `/etc/os-release`. Currently the wrapper script uses the latter. If It's "LinuxMint" then we do this the Mint-y way, if it's "Ubuntu" then we do this the Ubuntu-y way, otherwise, we error. I don't know if this logic is worth.

One option that we have is to do something like the following bash, but cleanly and in Ansible https://gist.github.com/kylelaker/1069416bc1955a48e020fad155a16dc3.

kylelaker:
The latest revision handles Ubuntu and Mint and should fail on all other distros

PR #113, opened 2018-02-20
Update JGrasp URL

### LINKED ISSUE #112: JGrasp 2.0.4_03 no longer available
The latest version of JGrasp is 2.0.4_04 and the old version is no longer available for download. This breaks CS149. `roles/jgrasp/vars/main.yml` should be updated with the updated download URL.

kylelaker:
Download the latest version of JGrasp. The 2.04_03 is no longer
available and breaks the playbook

Closes #112

### DIFF
--- a/roles/jgrasp/vars/main.yml
+++ b/roles/jgrasp/vars/main.yml
@@ -1,7 +1,7 @@
 ---
 # vars file for jgrasp
 jgrasp:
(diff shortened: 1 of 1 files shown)


kylelaker:
This will need to be backported to Sylvia. I will start that process once this is merged into master.

PR #120, opened 2018-03-04
wrapper: Fix style issues

kylelaker:
Don't shadow the quit() method. Even though we're not using it,
shadowing builtins is probably a bad idea. Additionally, since
on_close() never actually uses self, move it out of the window's class.


kylelaker:
That'd be ideal

kylelaker:
A qualified Python dev would be good too :rofl:

PR #123, opened 2018-03-22
Document Eclipse URL

kylelaker:
The comment adds information about the Eclipse download URL format to hopefully guard against links that contain a specific mirror or that don't directly link to the file.

Since this may change the future, I understand if this is not a desirable addition.


ripleymj:
If in doubt, I use the alternate link they provide: 
image

PR #132, opened 2018-04-25
Switch from gksudo to pkexec

kylelaker:


kylelaker:
Good point! Sorry, Richard. Ripley was just unlucky enough to be suggested by GItHub :)

roles/task-shortcuts/files/uug_ansible_wrapper.py
```diff
@@ -230,27 +230,39 @@ def show_about_dialog(self, _):
...
             logging.info("ansible-pull succeeded")
-        # 65280 should be the exit code if the gksudo dialog is dismissed
-        elif exit_status == 65280:
-            gksudo_err_msg = "Either an incorrect password was entered or" \
-                             " the password dialog was closed." \
-                             " Please try again"
+        # 126 should be the exit code if the pkexec dialog is dismissed and
+        # 127 is the exit code if authentication fails. All other exit codes
+        # come from the called application
+        elif exit_status == 126:
```
kylelaker: Currently, this stuff isn't working. It seems like `pkexec` when run on its own does exit with the correct codes (126 and 127); however, this keeps receiving the values 32226 and 32512 respectively under the described conditions.

roles/task-shortcuts/files/uug_ansible_wrapper.py
```diff
@@ -279,9 +291,9 @@ def on_run_clicked(self, _):
             invalid_branch(self)
             return
 
-        gksudo_msg = "<b>Enter your password to configure your VM</b>" \
-                     "\nTo configure your virtual machine, administrator" \
-                     " privileges are required."
+        # gksudo_msg = "<b>Enter your password to configure your VM</b>" \
```
kylelaker: These three lines should be removed, not just commented out

roles/task-shortcuts/tasks/main.yml
```diff
@@ -18,7 +18,17 @@
...
-- name: Install Ansible wrapper script
+- name: Install ansible-pull shell wrapper
+  copy:
+      src: ansible_pull_sh_wrapper
+      dest: "{{ bin_path }}"
+      mode: 0755
+- name: Install policykit-1 for Ansible wrapper
+  copy:
+      src: edu.jmu.uug.ansiblewrapper.policy
+      dest: /usr/share/policykit-1/actions/
```
kylelaker: `s/policykit/polkit/`

roles/task-shortcuts/tasks/main.yml
```diff
@@ -18,7 +18,17 @@
   file:
       path: "{{ bin_path }}"
       state: directory
-- name: Install Ansible wrapper script
+- name: Install ansible-pull shell wrapper
+  copy:
+      src: ansible_pull_sh_wrapper
+      dest: "{{ bin_path }}"
+      mode: 0755
+- name: Install policykit-1 for Ansible wrapper
```
kylelaker: Add a comment explaining why this has to go in /usr

kylelaker:
I force-pushed an updated version that is now ready for review. This has been tested on Ubuntu Bionic and the VM Sylvia release.

I believe the major issue left to discuss is whether to require the user to authenticate at all when using our wrapper. We could, in the policy, just set, `<allow_active>yes</allow_active>` instead of auth_admin; however, I am uncertain of the benefit of this. I don't love the idea of there being a way to arbitrarily run ansible-pull with root privileges.

kylelaker:
@ripleymj and I had a conversation about the shell wrapper for ansible-pull. That's now gone. This commit should be ready for review.

Sorry about all the stuff in regards to "Remember user's chosen roles." None of that is included in this commit but previously that commit's message included a reference to this pull request number.

I have tested this on Mint 18.3 and Ubuntu 18.04 and it worked as far as I could tell. It would be nice to have a second set (or more) of testing.

PR #137, opened 2018-06-18
Remember user's chosen roles

### LINKED ISSUE #135: Make the wrapper script remember roles
Each time the wrapper runs, ideally, the roles selected the last time the script was run will be selected automatically (but still be able to be unchecked). This probably will require a change in config format, possibly to JSON or YAML (and since so much of the project is YAML, it may be ideal).

I would like to see something like either of the following as a sample config.

```
---
git_url: "https://github.com/jmunixusers/cs-vm-build"
git_branch: sylvia
previous_roles:
  - common
  - cs101
```

or

```
{
"git_url": "https://github.com/jmunixusers/cs-vm-build",
"git_branch": "sylvia",
"previous_roles": [ "common", "cs101" ]
}
```

kylelaker:
There needs to be discussion about whether to remember every role a user has chosen over all runs of the script or only the roles the user chose the last time they ran. I prefer and have implemented the latter, but this can be adjusted to meet the requirements of the former.

Resolves #135

### DIFF
--- a/roles/task-shortcuts/files/uug_ansible_wrapper.py
+++ b/roles/task-shortcuts/files/uug_ansible_wrapper.py
@@ -3,9 +3,8 @@
    This tools creates a simple GUI for running ansible-pull with a
    predetermined set of tags. It displays the output from the ansible-pull
    command in a VTE within the GUI. It allows the user to override some things
-   in a configuration file (~/.config/vm_config). The branch to pull can be
-   overriden by setting FORCE_BRANCH and the URL to pull from can be overriden
-   with FORCE_GIT_URL
+   in a configuration file (~/.config/vm_config). The branch to pull and the
+   URL to pull from can be changed in the program's Settings.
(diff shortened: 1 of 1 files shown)


kylelaker:
Per a discussion with @ripleymj that did not make it to GitHub, it's been decided that the script should remember all the roles the user has ever chosen. 2de3947 is a commit that should implement that. Testing will be necessary and as such I am going to add that label.

kylelaker:
I've had to do a few updates after testing. As far as I can tell 1b3fb8f, which as of this comment, is the latest version, resolves all issues I've experienced.

Wording has been added to mention the fact that roles cannot be removed, only added & updated. All references to the old config are hopefully gone. Additional logging has been added. There have been some style fixes (and I hope to actually work more on style in the future).

I believe this PR is truly ready for review at this point.

ripleymj:

PR #138, opened 2018-06-29
Fix OEM role detecting Mint in Mint 19

kylelaker:
In Mint 19, ansible_distribution is "Linux Mint" rather than "Linuxmint"
as it was in previous versions of Mint. I have not checked whether this
is a change in Ansible or Mint; however, it is a second value that we
need to support.


ripleymj:

PR #139, opened 2018-06-29
Fix Mint 19 dependencies

kylelaker:
mint-artwork-gnome is no longer necessary and actually no longer exists
in the tara repos. The issues we saw previously where windows and
dialogs were missing some of their decorations no longer seems to occur.


ripleymj:

PR #140, opened 2018-06-29
Fix menubar spacing issue

kylelaker:
Ensure that the menubar remains flush to the top of the window instead
of having a 10-pixel border separating it from the top. Apply border
widths to the internal boxes instead of to the window itself.

This becomes especially noticeable with the Mint-Y theme used in Tara.

PR #141, opened 2018-06-29
Avoid needing a password for back-to-back runs

kylelaker:
This will allow users to only need to type their password probably once
every few runs. This essentially works like `sudo` where rather than
being prompted, the user can continue to execute it for a brief period
of time.


ripleymj:

PR #176, opened 2018-08-18
Add Ubuntu partner repo

### LINKED ISSUE #175: Investigate Mint codec problems
Per @ChrisMayfield,

> When I run "Install Multimedia Codes" from the menu, I get an error message that it failed and to fix broken packages first. So I ran "apt install mint-meta-codecs" from the terminal and got:
`The following packages have unmet dependencies:
 mint-meta-codecs : Depends: adobe-flashplugin but it is not installable
                    Depends: libavcodec-extra`

If we get this working, we can consider a backport, but I don't believe this should block Fall 2018 at this point.

kylelaker:
This is included by default in Mint; I am unsure why we weren't
including it previously. This will be required to install the media
codecs.

Resolves #175

### DIFF
--- a/roles/common/templates/mint.j2
+++ b/roles/common/templates/mint.j2
@@ -15,3 +15,6 @@ deb {{ mirror }} {{ ubuntu_release }}-security main restricted universe multiver
 deb {{ mirror }} {{ ansible_distribution_release }} main upstream import backport
 
 {% endfor %}
+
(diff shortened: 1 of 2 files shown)


ripleymj:
Since this seems simple enough, do you want to backport as well?

kylelaker:
This will be included in the mega backport

Issue #85, opened 2017-12-18
Missing robot packages


ripleymj:
These packages are missing in our image and present on the lab laptops. At this point, we don't know if they're necessary or just present. fonts-khmeros-core is a bad grep match.

```
$ diff -u our-pkgs.txt robot-pkgs.txt 
--- our-pkgs.txt	2017-12-17 21:25:20.529996425 -0500
+++ robot-pkgs.txt	2017-12-17 21:25:08.858224586 -0500
@@ -1,7 +1,17 @@
+fonts-khmeros-core				install
 ros-kinetic-actionlib				install
 ros-kinetic-actionlib-msgs			install
 ros-kinetic-actionlib-tutorials			install
+ros-kinetic-amcl				install
 ros-kinetic-angles				install
+ros-kinetic-ar-track-alvar			install
+ros-kinetic-ar-track-alvar-msgs			install
+ros-kinetic-ardrone-autonomy			install
+ros-kinetic-astra-camera			install
+ros-kinetic-astra-launch			install
+ros-kinetic-audio-common-msgs			install
+ros-kinetic-base-local-planner			install
+ros-kinetic-bfl					install
 ros-kinetic-bond				install
 ros-kinetic-bond-core				install
 ros-kinetic-bondcpp				install
@@ -9,8 +19,11 @@
 ros-kinetic-camera-calibration			install
 ros-kinetic-camera-calibration-parsers		install
 ros-kinetic-camera-info-manager			install
+ros-kinetic-capabilities			install
+ros-kinetic-carrot-planner			install
 ros-kinetic-catkin				install
 ros-kinetic-class-loader			install
+ros-kinetic-clear-costmap-recovery		install
 ros-kinetic-cmake-modules			install
 ros-kinetic-collada-parser			install
 ros-kinetic-collada-urdf			install
@@ -19,9 +32,16 @@
 ros-kinetic-compressed-depth-image-transport	install
 ros-kinetic-compressed-image-transport		install
 ros-kinetic-control-msgs			install
+ros-kinetic-controller-manager-msgs		install
+ros-kinetic-costmap-2d				install
 ros-kinetic-cpp-common				install
+ros-kinetic-create-dashboard			install
+ros-kinetic-create-description			install
+ros-kinetic-create-driver			install
+ros-kinetic-create-node				install
 ros-kinetic-cv-bridge				install
 ros-kinetic-depth-image-proc			install
+ros-kinetic-depthimage-to-laserscan		install
[...]

PR #194, opened 2018-09-11
Update CS354 packages

### LINKED ISSUE #192: Gazebo7 Package Error
The following error occurs for the `gazebo7` package:
```
2018-09-06 21:16:01,846 p=24242 u=root |  failed: [localhost] (item=[u'ros-kinetic-desktop-full', u'gazebo7', u'rviz', u'g++']) => {"changed": false, "item": ["ros-kinetic-desktop-full", "gazebo7", "rviz", "g++"], "msg": "No package matching 'gazebo7' is available"}
```

I also get this message when running `sudo apt-get install gazebo7` about there being a newer version with a package called `gazebo9`: 
```
Package gazebo7 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  gazebo9:i386 gazebo9

E: Package 'gazebo7' has no installation candidate
```

cfrome77:
ROS has been updated from kinetic to meldodic and the other two ROS packages were added per @kylelaker in #85. Melodic was not avaliable under xenial so I switched to bionic for the ROS packages link. Gazebo7 was updated to gazebo9 as mentioned in #192. This has been tested and all packages install correctly. The README was also updated to reflect the new package versions.

Closes #192

### DIFF
--- a/roles/robot-pkgs/tasks/main.yml
+++ b/roles/robot-pkgs/tasks/main.yml
@@ -24,8 +24,8 @@
     dest: '{{ global_profile_path }}'
     marker: "## {mark} ROS profile entries ##"
     block: |
-      source /opt/ros/kinetic/setup.bash
-      export TURTLEBOT_3D_SENSOR=kinect
+      source /opt/ros/melodic/setup.bash
+      export TURTLEBOT_3D_SENSOR=melodic
   notify:
   - Suggest restart
 - name: Init rosdep if missing
@@ -45,7 +45,7 @@
     group: '{{ item.gid }}'
   with_items: "{{ real_users }}"
 - name: Init Catkin workspace
-  shell: source /opt/ros/kinetic/setup.bash && catkin_make
+  shell: source /opt/ros/melodic/setup.bash && catkin_make
   args:
     executable: /bin/bash
     chdir: '{{ item.homedir }}/catkin_ws/'
--- a/roles/robot-pkgs/tasks/main.yml
+++ b/roles/robot-pkgs/tasks/main.yml
@@ -6,12 +6,12 @@
     id: 421C365BD9FF1F717815A3895523BAEEB01FA116
     state: present
 - name: Add ROS repository
-  apt_repository:
-    repo: "{{ item }}"
-    update_cache: no
-    filename: 'ros-latest'
-    state: present
-  with_items: "{{ ros_repos }}"
+  template:
+    src: ros-packages.j2
+    dest: /etc/apt/sources.list.d/ros-packages.list
+    owner: root
+    group: root
+    mode: 0644
 - name: Refresh apt cache
   apt:
     update_cache: yes
@@ -24,8 +24,8 @@
     dest: '{{ global_profile_path }}'
     marker: "## {mark} ROS profile entries ##"
     block: |
-      source /opt/ros/melodic/setup.bash
-      export TURTLEBOT_3D_SENSOR=melodic
+      source /opt/ros/{{ros_release}}/setup.bash
+      export TURTLEBOT_3D_SENSOR={{ros_release}}
   notify:
   - Suggest restart
 - name: Init rosdep if missing
@@ -45,7 +45,7 @@
     group: '{{ item.gid }}'
   with_items: "{{ real_users }}"
 - name: Init Catkin workspace
-  shell: source /opt/ros/melodic/setup.bash && catkin_make
+  shell: source /opt/ros/{{ros_release}}/setup.bash && catkin_make
   args:
     executable: /bin/bash
     chdir: '{{ item.homedir }}/catkin_ws/'
--- a/roles/robot-pkgs/tasks/main.yml
+++ b/roles/robot-pkgs/tasks/main.yml
@@ -1,5 +1,9 @@
 ---
 # tasks file for robot-pkgs
+- name: Install fonts-khmeros-core package
+  apt: name={{ item }} state=latest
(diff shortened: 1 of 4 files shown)


ripleymj:
Thank you for putting this together! I'm a bit torn on how to proceed at the moment. On one hand, we know the current content is broken and nobody depends on it this semester. On the other hand, the robotics lab is currently missing after the EnGeo renovation and the robotics club is trying to get itself back together, so we have no means to test these changes. I'll try to do an actual review in the next few days, but may hold merging until we figure out where we're going.

roles/robot-pkgs/tasks/main.yml
```diff
@@ -1,5 +1,9 @@
 ---
 # tasks file for robot-pkgs
+- name: Install fonts-khmeros-core package
+  apt: name={{ item }} state=latest
+  with_items:
+    - fonts-khmeros-core
```
kylelaker: We actually don't need `fonts-khmeros-core`. I think it snuck into the list because when you `grep ros`, it ends up making it in. Sorry about that -- do you mind removing this bit?

roles/robot-pkgs/vars/main.yml
```diff
@@ -3,12 +3,14 @@
 
 # Keep UMD first to prefer downloads from there
 ros_repos:
-  - deb http://mirror.umd.edu/packages.ros.org/ros/ubuntu xenial main
-  - deb http://packages.ros.org/ros/ubuntu xenial main
+  - deb http://mirror.umd.edu/packages.ros.org/ros/ubuntu bionic main
+  - deb http://packages.ros.org/ros/ubuntu bionic main
```
kylelaker: There should be an `ubuntu_release` variable that gets registered in `common`. Can we change this to use a Jinja2 template and then put this into `/etc/apt/sources.list.d/ros-packages.list`? The template would go in `roles/robot-pkgs/templates/ros-packages.list.j2` and the files in `roles/common/templates/*.list` would be a good template (but clearly only need these two repos, not the full suite of Ubuntu/Mint repos).

I'd also like to see the mirrors enumerated in the `vars/main.yml` file. So ideally, something like:

```
{%for mirror in ros_mirrors %}
deb {{ mirror_url }} {{ ubuntu_release }} main
{% endfor %}
```

but the syntax, etc will need to be verified/tested.

This would also require changes to `tasks/main.yml` to no longer use the `apt_repository` module and instead use the `template` module.

roles/robot-pkgs/tasks/main.yml
```diff
@@ -24,8 +28,8 @@
     dest: '{{ global_profile_path }}'
     marker: "## {mark} ROS profile entries ##"
     block: |
-      source /opt/ros/kinetic/setup.bash
-      export TURTLEBOT_3D_SENSOR=kinect
+      source /opt/ros/melodic/setup.bash
```
kylelaker: Can we add a variable called `ros_release` and set it to `melodic`?

roles/robot-pkgs/tasks/main.yml
```diff
@@ -45,7 +49,7 @@
     group: '{{ item.gid }}'
   with_items: "{{ real_users }}"
 - name: Init Catkin workspace
-  shell: source /opt/ros/kinetic/setup.bash && catkin_make
+  shell: source /opt/ros/melodic/setup.bash && catkin_make
```
kylelaker: See comment on `ros_release` variable.

roles/robot-pkgs/vars/main.yml
```diff
@@ -3,12 +3,14 @@
...
 
 ros_packages:
-    - ros-kinetic-desktop-full
-    - gazebo7
+    - ros-melodic-desktop-full
```
kylelaker: See comment on `ros_release` variable.

kylelaker:
Hopefully this will also close #85.

kylelaker:
Targeting to Spring 2019 which is where the other robot-related issues are. I'd like to see this in long before then, but I also want to see this count towards that milestone :)

roles/robot-pkgs/tasks/main.yml
cfrome77: removed with commit 5bb24a2

roles/robot-pkgs/tasks/main.yml
cfrome77: removed with commit 5bb24a2

roles/robot-pkgs/vars/main.yml
cfrome77: This will definitely eliminate having to change the release name every time. See  commit 91b364e for the Jinja2 template

roles/robot-pkgs/tasks/main.yml
cfrome77: Again, a great idea so we only have to type the release name once. See commit f8e93cb.

roles/robot-pkgs/tasks/main.yml
cfrome77: See commit 96fd33a

roles/robot-pkgs/vars/main.yml
cfrome77: See commit f8e93cb

cfrome77:
@kylelaker I liked your suggestions, as it will make updating for new releases easier in the future. I went ahead and implement the changes. I commented the commit hash that reflects the change under each suggestion.

PR #204, opened 2018-10-21
Fix timeshift

### LINKED ISSUE #200: Timeshift potentially fills hard disk 100%
The new Timeshift feature that ships with Mint 19 aggressively uses disk space. It's not turned on by default, but the Update Manager "warns" (nags) users to set up snapshots. After some time (several weeks?) the /timeshift/snapshots directory takes about 10 GB. If the disk usage reaches 100%, the VM only boots to tty1. Deleting the contents of /timeshift/snapshots immediately fixes the problem. I've reported this issue upstream: https://bugs.launchpad.net/linuxmint/+bug/1738065

As a workaround (to prevent this issue in future semesters), you might want to add an Ansible task that unchecks "Show a warning if the system snapshots are not set up" in the Update Manager. Or maybe it would be better to remove Timeshift altogether. There might be a way to configure /etc/timeshift.json to avoid this issue.

showalrm:
This removes the Timeshift package on oem runs and removes the warning in the Mint update manager.

Closes #200

### DIFF
--- a/roles/common/handlers/main.yml
+++ b/roles/common/handlers/main.yml
@@ -14,3 +14,9 @@
   become: yes
   become_user: "{{ item.user }}"
   with_items: "{{ real_users }}"
+- name: Restart update manager
+  shell:
+    nohup mintupdate-launcher
+  become: yes
+  become_user: "{{ item.user }}"
+  loop: "{{ real_users }}"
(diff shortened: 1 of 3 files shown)


kylelaker:
@ripleymj does this need to be backported to tara?

PR #207, opened 2018-10-23
Fix looping style

### LINKED ISSUE #203: Adopt Ansible 2.5-style looping
I stumbled across an Ansible document that claims our typical `with_items` style loops have been deprecated in favor of the `loop` keyword: https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#migrating-from-with-x-to-loop

At a minimum, we should probably adopt this style as new code goes in, with an eye to update more places as appropriate.

coledn:
with_items were all changed to loop. Amended less than @showalrm :D.

Closes #203

### DIFF
--- a/roles/adv-prog-pkgs/tasks/main.yml
+++ b/roles/adv-prog-pkgs/tasks/main.yml
@@ -1,27 +1,10 @@
 ---
 # tasks file for adv-prog-pkgs
 - name: Install advanced language packages
-  apt: name={{ item }} state=latest
-  loop:
-    - build-essential
-    - g++
-    - gcc
-    - gdb
-    - logisim
-    - valgrind
+  apt: name={{ adv_prog_pkgs_adv_lang }} state=latest
+
 - name: Install source control packages
-  apt: name={{ item }} state=latest
-  loop:
-    - git
-    - gitg
-    - mercurial
-    - meld
+  apt: name={{   adv_prog_pkgs_souce_control }} state=latest
+
 - name: Install code editor packages
-  apt: name={{ item }} state=latest
-  loop:
-    - astyle
-    - bvi
-    - check
-    - gedit
-    - indent
-    - nano
+  apt: name={{ adv_prog_pkgs_code_editor }} state=latest
--- a/roles/adv-prog-pkgs/tasks/main.yml
+++ b/roles/adv-prog-pkgs/tasks/main.yml
@@ -4,7 +4,7 @@
   apt: name={{ adv_prog_pkgs_adv_lang }} state=latest
 
 - name: Install source control packages
-  apt: name={{   adv_prog_pkgs_souce_control }} state=latest
+  apt: name={{ adv_prog_pkgs_souce_control }} state=latest
 
 - name: Install code editor packages
   apt: name={{ adv_prog_pkgs_code_editor }} state=latest
--- a/roles/adv-prog-pkgs/tasks/main.yml
+++ b/roles/adv-prog-pkgs/tasks/main.yml
@@ -2,7 +2,7 @@
 # tasks file for adv-prog-pkgs
 - name: Install advanced language packages
   apt: name={{ item }} state=latest
-  with_items:
+  loop:
     - build-essential
     - g++
     - gcc
@@ -11,14 +11,14 @@
     - valgrind
 - name: Install source control packages
   apt: name={{ item }} state=latest
-  with_items:
+  loop:
     - git
     - gitg
     - mercurial
     - meld
 - name: Install code editor packages
   apt: name={{ item }} state=latest
-  with_items:
+  loop:
     - astyle
     - bvi
     - check
--- a/roles/adv-prog-pkgs/vars/main.yml
+++ b/roles/adv-prog-pkgs/vars/main.yml
@@ -1,2 +1,23 @@
 ---
 # vars file for adv-prog-pkgs
+adv_prog_pkgs_adv_lang:
+  - build-essential
+  - g++
+  - gcc
+  - gdb
+  - logisim
+  - valgrind
+
+adv_prog_pkgs_souce_control:
+  - git
+  - gitg
+  - mercurial
+  - meld
+
+adv_prog_pkgs_code_editor:
+  - astyle
+  - bvi
+  - check
+  - gedit
+  - indent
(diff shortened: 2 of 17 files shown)


ripleymj:
I can't believe I missed this behavior of apt, especially after we spent so much time trying to understand it initially. I don't think the `loop | flatten` syntax will help, as that seems to just flatten nested lists into a single-level list.

While we could probably change it to:
```
- name: Install packages
  apt:
    state: present
    name:
      - package1
      - package2
```
this might be the time to look at which of these should be moved to a named list in `vars/main.yml` and if we need to adopt a naming convention to prevent them from overwriting each other, as I believe they're all global. Perhaps just prefixing them with the role name.

roles/adv-prog-pkgs/tasks/main.yml
```diff
@@ -1,27 +1,10 @@
...
+  apt: name={{ adv_prog_pkgs_adv_lang }} state=latest
+
 - name: Install source control packages
-  apt: name={{ item }} state=latest
-  with_items:
-    - git
-    - gitg
-    - mercurial
-    - meld
+  apt: name={{   adv_prog_pkgs_souce_control }} state=latest
```
kylelaker: There are a few extra spaces here. Can those be removed?

```suggestion
  apt: name={{ adv_prog_pkgs_souce_control }} state=latest
```

roles/basic-prog-pkgs/vars/main.yml
```diff
@@ -1,2 +1,18 @@
 ---
-# vars file for basic-prog-pkgs
\ No newline at end of file
+# vars file for basic-prog-pkgs
+basic_prog_pkgs_intro_development:
+ -artha
```
kylelaker: Can you add a space after the `-`?

```suggestion
 - artha
```

roles/common/tasks/main.yml
```diff
@@ -34,6 +34,7 @@
 - name: Get users
   getent:
     database: passwd
+ # fix this after Ansible 2.6
```
kylelaker: Any chance we can elaborate on exactly how we should fix it and what needs to be done after Ansible 2.6? Maybe @ripleymj can provide some insight here?

roles/adv-prog-pkgs/tasks/main.yml
```diff
@@ -1,27 +1,10 @@
...
 - name: Install advanced language packages
-  apt: name={{ item }} state=latest
-  with_items:
-    - build-essential
-    - g++
-    - gcc
-    - gdb
-    - logisim
-    - valgrind
+  apt: name={{ adv_prog_pkgs_adv_lang }} state=latest
```
kylelaker: Nice! I like that these have become one-line entries in the config files! Thanks!

roles/common/tasks/main.yml
showalrm: It should be changed to `loop: "{{ getent_passwd|dict2items }}"` once mint uses Ansible 2.6
[as seen here.](https://docs.ansible.com/ansible/2.6/user_guide/playbooks_loops.html#with-dict)

Ansible 2.5 just doesn't support this type of styling yet.

ripleymj: ^^this

kylelaker: Nice! Thanks!

kylelaker:
```suggestion
 # fix this after Ansible 2.6, use loop: "{{ getent_passwd|dict2items }}"
```

ripleymj:
@coledn please comment that you're finished and ready for review when you're finished and ready for review.

coledn:
I believe I am done. Please look it over in case I made some mistakes.

ripleymj:
LPT: hit Submit Review, don't just leave it as pending for an hour :(

roles/adv-prog-pkgs/vars/main.yml
```diff
@@ -1,2 +1,23 @@
...
+  - mercurial
+  - meld
+
+adv_prog_pkgs_code_editor:
+  - astyle
+  - bvi
+  - check
+  - gedit
+  - indent
+  - nano
```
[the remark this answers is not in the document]
kylelaker: For a first commit, definitely not :) I'd be happy to see it added, but I plan to do a newline cleanup in a few weeks anyway.

roles/adv-prog-pkgs/vars/main.yml
```diff
@@ -1,2 +1,24 @@
...
+
```
kylelaker: Hate to be nitpicky, but can you remove the whitespace on this line? Otherwise there isn't actually a newline at the end of the file (`'\n'`) it's a space (`' '`).

```suggestion

```

ripleymj:
I think that takes care of everything I've been able to spot. Unless somebody else confirms they've done this, I do want to let Ansible run the entire playbook before approving this for merging, especially since it touches almost everything.

Great job though, I can't believe you accepted our bait to take on this project. :)

kylelaker:
I can confirm that the playbook ran without any issues for me. Definitely doesn't hurt to run it on another VM, though. I also didn't test the OEM role, but it looks fine.

@coledn this is fantastic work! Thanks for your patience and for doing such a great job. I looked through all our pull requests and this had the highest number of comments and discussion of any of our PRs so far :1st_place_medal:

This was a much more significant change than @ripleymj or I expected and it's awesome you got it knocked out so quickly. Thank you!

coledn:
I ran a test on the VM myself as well and it had no issues.

ripleymj:
I've hit an error twice now running the CS354 content: https://gist.github.com/ripleymj/f0da484c92165fcc2f736a54f2e5f3a4

I doubt it's related to your changes, but I'd like to try to understand it better before merging.

coledn:
Just fixed it, try to see if you get errors now.

coledn:
@ripleymj

ripleymj:
I was able to run all class targets here, and the robot error didn't show up.

@showalrm, you added yourself as a reviewer - don't disappoint @coledn now.

showalrm:
I think that was an unintended consequence of joining the conversation earlier.

The changes to the code look good and when I did testing everything went smoothly. Great job @coledn

Issue #206, opened 2018-10-22
Add support for 361


Advill:
Some 361 required C libraries are not included in the image. I don't have the specific libraries yet, will be testing later this week. Will mark down here what is missing specifically.

Advill:
I believe that this can be fixed simply by running the `adv-prog-pkgs` role, I've implemented this [here in my fork](https://github.com/advill/cs-vm-build), but am holding the pull request as I have not tested it yet, nor added it to the GUI.

ripleymj:
I'm a little confused by what git's trying to tell me. Do you want to rename `adv-prog-pkgs` or create a new subset role? If you do that, do we trigger `c-programming-essentials` for both 261 and 361, and then just `c-programming-essentials` for 361? If you're creating a new role, use `ansible-galaxy init` to get the skeleton framework.

We've got a couple other projects going on to reorganize `common` and `task-shortcuts`, as well as cleaning up the looping syntax throughout the project. I'd just like to understand where you're headed to avoid sadness of people tripping over each other.

You may also want to start down a branch rather than hacking directly in `master`.

Advill:
At first I thought i needed to add a new role, but after checking I realized all packages should be installed by `adv-prog-pkgs`, so I've simply added a new tag to that roll in the `local.yml` file. This shouldn't step on anyone, at least yet, as this is the only change I've made as of yet, and hopefully the only change I need to make.

ripleymj:
Ahh, gotcha! Yep, start with `local.yml` and then `task-shortcuts/files/uug_ansible_wrapper.py` if you're feeling adventurous.

kylelaker:
So, if 361 is just going to install the same things as CS 261 (minus `y86`), I'd almost rather we just change the wrapper to say, `CS 261/361`. If we can find some meaningful difference in the package requirements, I am okay with splitting them off into separate tags.

Advill:
That's true. There shouldn't be anything missing after the install of `build-dependencies` and `g++` right? I'll pull the VM and take a look to be sure, but considering there's nothing extreme I'll close this.

ripleymj:
I'm ok with changing the label in the wrapper to indicate support for CS 361. Feel free to leave this issue open to track that.

kylelaker:
[...]

PR #208, opened 2018-10-24
Add CS-361 to 261 option on install

Advill:
Closes jmunixusers/cs-vm-build#206

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ apt will require more extensive modifications.
 ### CS159
 * Eclipse
 
-### CS261
+### CS261 / CS361
 * Mint packages (adv-prog-pkgs role)
   * Compilers and interpreters
     * build-essential
@@ -90,28 +90,6 @@ apt will require more extensive modifications.
 * ROS Kinetic
 * Rosdep/catkin initialization
 * Rviz
-
-### CS361
-* Mint packages (adv-prog-pkgs role)
-  * Compilers and interpreters
-    * build-essential
-    * g++
-    * gcc
-    * gdb
-    * logisim
-    * valgrind
-  * Code editors
-    * astyle
-    * bvi
-    * check
-    * gedit
-    * indent
-    * nano
-  * Source control
-    * git
-    * gitg
-    * mercurial
-    * meld
 
 ## License
 


ripleymj:
Whatever direction we decide to go, I think we need to put merging 361 on hold pending the outcome of @coledn's rewrite.

kylelaker:
Welp... I feel like a jerk. I added a comment on #206 perhaps a little late explaining that we probably should split the tags for these courses. Would you mind looking at that comment and either giving your thoughts here?

Basically, my concerns are centered around 'What if we need to split 261 and 361 in the future' and I hadn't considered that previously when suggesting just modifying the script.

Advill:
No worries. I can get that done fairly quickly tomorrow if we are more or less set on that option. None of my changes should interfere with @coledn 's changes, as I'm just messing with tags, supposedly localized entirely within the `local.yml` file and perhaps the readme. If it is preferred that I wait until uug to figure that out that is fine with me.

kylelaker:
I'm okay if you start work on it; I just feel like there might be some questions left to answer such as:

 - 361 requires (at least) all of 261 minus y86, right?
 - It will also eventually require its own role since it needs it's own packages?
    - If so, what do we want to call that new role? We've already taken `adv-prog-pkgs` for 261.
    - If so, it should be created with `ansible-galaxy init <role_name>` in the `roles/` dir
 - Do we know what packages 361 needs?

Advill:
To answer your questions in order as far as I know at the moment:
- Yes, even some of the packages in `adv-prog-pkgs` might be extraneous, but removing them likely isn't worth it.
- I'm not sure, for where we are right now in Aboutabl's class, it doesn't appear that we've needed anything else. I intend on ensuring that this is true, but haven't had the time to do so yet.
    - I think something to do with `multithreaded-prog-pkgs` could do
    - Yes, I've got this downloaded now, ansible was being weird about installing through pacman at first.
- As the year goes on, I'd be happy to add any packages that come up, but knowing Aboutabl I wouldn't expect anything too far outside the basic c libraries.

I'm certainly open to changes if we find different answers, my plan will be to implement things like this right now as it won't be that difficult of a change if we find things are different.

kylelaker:
That all sounds good! I think I'm fine with having a basic 361 tag set up and then modifying as needed. Let's not create the new role until it's needed, though

Advill:
closed in favor of #209

PR #210, opened 2018-10-25
Fix terminal scrolling in wrapper

### LINKED ISSUE #190: Cannot scroll previous output in GUI
It would be nice if you could scroll back the output in the "JMU CS VM Configuration" tool, especially if there are errors or warnings. Right now there's no scroll bar, and wheel / two-finger scrolling doesn't work.

kylelaker:
Currently the terminal prevents users from inputting things (such as ^C)
by using the set_sensitive(False) function call. Unfortunately, this
also disables scrolling. The set_input_enabled(False) call may make
this better. This also adds a call to set_scroll_on_output to ensure
that any text written to the terminal is seen to avoid situations where
a user scrolls and then doesn't see any output so it looks like Playbook
is stuck.

Closes #190

### DIFF
--- a/roles/task-shortcuts/files/uug_ansible_wrapper.py
+++ b/roles/task-shortcuts/files/uug_ansible_wrapper.py
@@ -152,7 +152,12 @@ class AnsibleWrapperWindow(Gtk.Window):
 
         # Add the terminal to the window
         self.terminal = Vte.Terminal()
-        self.terminal.set_sensitive(False)
+        # Prevent the user from entering text or ^C
+        self.terminal.set_input_enabled(False)
+        # Ensure that if text is written, the user sees it
+        self.terminal.set_scroll_on_output(True)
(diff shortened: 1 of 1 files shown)


kylelaker:
I don't disagree; however, I believe that would require wrapping the VTE in a container and then finding a way to grow the VTE widget as the content grows. I would need to read a bit more about these specific widgets before that'd be possible.

I think this gets us back to where we were initially with this widget before `set_sensitive(False)` was used but with scrolling re-enabled.

I'll leave it to you as a reviewer to decide whether scroll bars should be implemented as part of this PR. It will take time but is almost certainly doable.

ripleymj:
If it's not a `setScrollable(True)`, then let's take this as it is. I think we mentioned adding a "Open log directory" menu option to the wrapper at some point, which will likely be a better way to inspect this output.

PR #215, opened 2018-10-25
Cleanup style within the config tool script

kylelaker:
This makes style across the script signficantly more uniform. yapf was
used to get started on the work but a bit of it needed to be cleaned up
manually; however, in the future, the following yapf config should be
fine.

    [style]
    based_on_style = pep8
    dedent_closing_brackets = true
    split_complex_comprehension = true


kylelaker:
@ripleymj Rebase complete

PR #217, opened 2018-10-26
Eclipse timeout

### LINKED ISSUE #191: Eclipse download timeout
Several people reported Eclipse bundle fetch timed out. Let's see if we can make this better.

Advill:
closes #191 

Have script set up to timeout after 30 seconds, as well as a backup to try to pull from the vt mirror.

### DIFF
--- a/roles/eclipse/tasks/main.yml
+++ b/roles/eclipse/tasks/main.yml
@@ -13,15 +13,15 @@
         dest: '{{ eclipse.zip }}'
         timeout: 30
         force: yes
-		register: eclout
-		ignore_errors: true
-	- name: Fetch Eclipse bundle from backup server
-	  get_url:
-		url: '{{ eclipse.urlbck }}'
-		dest: '{{ eclipse.zip }}'
-		timeout: 30
-		force: yes
-		when: "'FAILED' in eclout.stderr"
+        register: eclout
+        ignore_errors: true
+    - name: Fetch Eclipse bundle from backup server
+      get_url:
+        url: '{{ eclipse.urlbck }}'
+        dest: '{{ eclipse.zip }}'
+        timeout: 30
+        force: yes
+        when: "'FAILED' in eclout.stderr"
     - name: Remove old Eclipse directory
       file:
         path: '{{ eclipse.install_path }}'
--- a/roles/eclipse/tasks/main.yml
+++ b/roles/eclipse/tasks/main.yml
@@ -12,8 +12,16 @@
         url: '{{ eclipse.url }}'
         dest: '{{ eclipse.zip }}'
         timeout: 30
-        retries: 10
         force: yes
+		register: eclout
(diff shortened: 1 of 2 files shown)


roles/eclipse/vars/main.yml
```diff
@@ -4,6 +4,7 @@ eclipse:
   # Ensure the URL does not specify a mirror and that &r=1 is at the end, which
   # directly links to the file and not the web page with a download button
   url: 'https://www.eclipse.org/downloads/download.php?file=/technology/epp/downloads/release/photon/R/eclipse-java-photon-R-linux-gtk-x86_64.tar.gz&r=1'
+  urlbck: 'https://www.eclipse.org/downloads/download.php?file=/oomph/epp/photon/R/eclipse-inst-linux64.tar.gz&mirror_id=504'
```
kylelaker: We need the URL for Eclipse for Java Developers, not the installer, so it should have a URL with `eclipse-java-photon-R-linux-gtk-x86_64.tar.gz` somewhere in it.

Additionally, can this variable use a longer name without an abbreviation like `backup_url` or `url_backup`?

roles/eclipse/tasks/main.yml
```diff
@@ -11,7 +11,17 @@
       get_url:
         url: '{{ eclipse.url }}'
         dest: '{{ eclipse.zip }}'
+        timeout: 30
         force: yes
+      register: eclout
+      ignore_errors: true
+    - name: Fetch Eclipse bundle from backup server
+      get_url:
+        url: '{{ eclipse.urlbck }}'
```
kylelaker: See comment on naming this variable

roles/eclipse/vars/main.yml
Advill: ah yes, I'll get that set up

roles/eclipse/tasks/main.yml
Advill: Will do

kylelaker:
@Advill I believe the backup URL we'd want to use is:

http://mirror.cc.vt.edu/pub/eclipse/technology/epp/downloads/release/photon/R/eclipse-java-photon-R-linux-gtk-x86_64.tar.gz

ripleymj:
I keep seeing references to "2018-09 R", instead of Photon. Can you see if we need to upgrade this while you're in there tinkering?

kylelaker:
@ripleymj It seems the Eclipse has moved to a new release cycle. We need to determine which release we want to include. That may be beyond the scope of this PR.

https://wiki.eclipse.org/Simultaneous_Release

roles/eclipse/vars/main.yml
ripleymj: In the looping rewrite, I was encouraging prefixing variable names with the role name, as they're all global.

Issue #220, opened 2018-10-31
jGRASP unavailable


kylelaker:
jGRASP 2.0.5_03 has been released which means that the version we are installing is no longer available. Users who install jGRASP for the first time or whose zip file have been corrupted or deleted will have their Ansible runs fail.

ripleymj:
@showalrm or @coledn do you want to take this one? You'll need to update the URL and SHA1 hash in the jGRASP vars file, and then prepare a backport to tara. Please self-assign the issue if you're interested.

kylelaker:
Issues has been assigned. Removing 'help wanted' tag. Thanks for taking this @coledn! Let us know if you have any questions!

kylelaker:
Despite @coledn's message in #222, this issue was not automatically closed. Closing since both #221 and #222 fixed this in `master` and `tara` respectively.

PR #221, opened 2018-11-01
Fix jGRASP hash and url

coledn:


kylelaker:
Thanks! We'll close #220 when we merge the backport!

PR #223, opened 2018-11-01
Fix branch handling for upgrades

### LINKED ISSUE #211: Improve branch handling after upgrade
Handling Mint upgrades is a huge conversation for us, but I'd like to propose one piece.

In our wrapper, could a check be added to warn (or offer to replace) the currently chosen branch when:
- url is set to github.com/jmunixusers/cs-vm-build
- branch != os-release

The branch would then be set to the value from os-release, perhaps after a git probe to see if it exists, and master if that fails. Checking these two conditions should prevent unnecessary warnings for us when developing, and attempt to improve Ansible handling for people who choose to upgrade their VM.

It is not a goal of this issue to attempt to upgrade VMs.

cfrome77:
This code warns the user about using an outdated branch or about using master, and suggests the correct branch to use based on their current Mint release version. It does not hinder the code from running, only warns the user before running the code. 

Fixes parts of #211

### DIFF
--- a/roles/common/files/uug_ansible_wrapper.py
+++ b/roles/common/files/uug_ansible_wrapper.py
@@ -666,7 +666,7 @@ def valdiate_config_settings(self):
             self, Gtk.MessageType.WARNING, Gtk.ButtonsType.OK, header, warning_prompt
         )
             
-    return warning_bool
+    return warning_prompt is None
 
 
 def validate_branch():
--- a/roles/common/files/uug_ansible_wrapper.py
+++ b/roles/common/files/uug_ansible_wrapper.py
@@ -23,6 +23,8 @@ from gi.repository import Gtk, Vte
 from gi.repository import GLib
 from gi.repository import GdkPixbuf
 
+DEFAULT_GIT_REMOTE = "https://github.com/jmunixusers/cs-vm-build"
+
 # If no tags are passed to ansible-pull, all of them will be run and I am
 # uncertain of the outcome of passing -t with no tags. To avoid this, always
 # ensure that common is run by adding it to this list and disabling the
@@ -47,7 +49,6 @@ USER_CONFIG = {
 }
 NAME = "JMU CS VM Configuration"
 VERSION = "Spring 2019"
-DEFAULT_GIT_REMOTE = "https://github.com/jmunixusers/cs-vm-build"
 
 def main():
     """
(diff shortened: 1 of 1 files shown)


roles/common/files/uug_ansible_wrapper.py
```diff
@@ -646,6 +652,42 @@ def unable_to_detect_branch():
...
+        "You are currently an unstable development branch (master) of the configuration tool. "
+        "You should consider switching to the release branch for your Linux Mint version. "
+        "The correct configuration settings for your release version of Linux Mint is:" 
+        "\nRelease: %(0)s and URL: %(1)s" % {
+            '0': get_distro_release_name(),
+            '1': USER_CONFIG['git_url']
+        }
+    )
+    show_dialog(parent, Gtk.MessageType.WARNING, Gtk.ButtonsType.OK,
+                           "Current branch is development branch master", warning_prompt)
```
kylelaker: I'd like to see us become a little more consistent around syntax more like the following when we wrap lines:

```
func_name(
    param1,
    param2,
    param3,
)
```

OR:

```
func_name(
    param1, param2, param3,
    param4, param5, param6,
)
```

Whichever feels more appropriate in the moment. I think maybe the second would be best here? I'll leave it up to you to decide.

Anyway, note the line break after `(` and before `)` and how the first character of each parameter lines up.

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -388,6 +388,14 @@ def on_run_clicked(self, _):
         if not validate_branch():
             invalid_branch(self)
             return
+        
+        if (get_distro_release_name() > USER_CONFIG['git_branch']) and (USER_CONFIG['git_url'] == "https://github.com/jmunixusers/cs-vm-build") and (USER_CONFIG['git_branch'] != "master"):
```
kylelaker: Am I misunderstanding this, or could the two if bodies be collapsed to a single check?

```python
if (get_distro_release_name() != USER_CONFIG['git_branch'] and (USER_CONFIG['git_url'] == "https://github.com/jmunixusers/cs-vm-build") and (USER_CONFIG['git_branch'] != "master")
```

We could even remove the `USER_CONFIG['git_branch'] != 'master'` if the conditional on Line 397 were done first.

kylelaker: Can the URL be moved to a global constant called `DEFAULT_GIT_REMOTE` or something similar?

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -388,6 +388,14 @@ def on_run_clicked(self, _):
...
+            warn_wrong_release_branch(self)
+        elif (get_distro_release_name() < USER_CONFIG['git_branch']) and (USER_CONFIG['git_url'] == "https://github.com/jmunixusers/cs-vm-build") and (USER_CONFIG['git_branch'] != "master"):
+            warn_wrong_release_branch(self)
+
+        if (USER_CONFIG['git_branch'] == "master") and (USER_CONFIG['git_url'] == "https://github.com/jmunixusers/cs-vm-build"):
```
kylelaker: See above comment on global constant & moving this check to be first

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -388,6 +388,14 @@ def on_run_clicked(self, _):
         if not validate_branch():
             invalid_branch(self)
             return
+        
+        if (get_distro_release_name() > USER_CONFIG['git_branch']) and (USER_CONFIG['git_url'] == "https://github.com/jmunixusers/cs-vm-build") and (USER_CONFIG['git_branch'] != "master"):
+            warn_wrong_release_branch(self)
+        elif (get_distro_release_name() < USER_CONFIG['git_branch']) and (USER_CONFIG['git_url'] == "https://github.com/jmunixusers/cs-vm-build") and (USER_CONFIG['git_branch'] != "master"):
```
kylelaker: See above comment on collapsing if checks and global constant for URL

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -596,7 +604,6 @@ def validate_branch():
 
     return USER_CONFIG['git_branch'] in ls_remote_output
 
-
```
kylelaker: If this is not inside a class, I'd like to keep this empty line. I'd prefer to do:
 - one line between functions inside a class
 - two lines between all global-level functions, between a function and a constant, and between a constant and imports.

As an example of the latter:

```python
import os
import sys


GLOBAL_VAR_A = 1
GLOBAL_VAR_B = 2


class Test(object):
    def __init__(self):
        pass

    def _foo(self):
        pass


def bar():
    pass



def baz():
    pass
```

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -646,6 +652,42 @@ def unable_to_detect_branch():
...
+def warn_wrong_release_branch(parent):
+    """
+    Displays a dialog to warn the user that their current branch in the configuration tool does not
+    match their os rlease version.
+    """
+
+    logging.info("The chosen branch is < OS releae. Warning user about switching branches")
+    warning_prompt = (
+        "You are using a version of the configuration tool meant for a different release of Linux Mint. " 
+        "You should consider upgrading braches. The correct configuration settings for your release version of Linux Mint is:"
```
kylelaker: Minor wording suggestions
```suggestion
        "You should consider upgrading branches. The recommended configuration settings for your release version of Linux Mint are:"
```

kylelaker: Can we make sure this wraps at approx 78-80 chars?

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -578,12 +574,106 @@ def get_distro_release_name():
...
+            warning_prompt = (
+                "You are currently on an unstable development branch (master) of the configuration tool. "
+                "You should consider switching to the release branch for your Linux Mint version. "
+                "The recommended configuration settings for your release version of Linux Mint are:" 
+                "\nRelease: %(0)s and URL: %(1)s" % {
+                    '0': get_distro_release_name(),
+                    '1': DEFAULT_GIT_REMOTE
+                }
+            )
+            header = "using testing branch master"
```
kylelaker:
```suggestion
            header = "Using testing branch master"
```

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -578,12 +574,106 @@ def get_distro_release_name():
...
+            )
+            header = "OS not supported and branch not found"
+            warning_bool = False
+        
+    if header and warning_prompt:
+        show_dialog(
+            self, Gtk.MessageType.WARNING, Gtk.ButtonsType.OK, header, warning_prompt
+        )
+            
+    return warning_bool
```
kylelaker:
```suggestion
    return warning_prompt is None
```

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -388,7 +388,10 @@ def on_run_clicked(self, _):
         if not validate_branch():
             invalid_branch(self)
             return
-
+        
+        if not validate_config_settings(self):
```
kylelaker: Since this is supposed to just be a warning, let's still continue even if it fails.

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -578,12 +574,106 @@ def get_distro_release_name():
...
+    output = subprocess.run(
+        ["/usr/bin/git", "ls-remote", '--heads', USER_CONFIG['git_url']],
+        stdout=subprocess.PIPE
+    )
+    
+    ls_remote = output.stdout.decode("utf-8")
+    
+    header = None
+    warning_prompt = None
+    warning_bool = None
```
kylelaker: Let's remove this variable

roles/common/files/uug_ansible_wrapper.py
kylelaker: If `warning_prompt is not None` then there was a warning set and so it should have the same value as `warning_bool`.

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -578,12 +574,101 @@ def get_distro_release_name():
     return release
 
 
+def valdiate_config_settings(self):
```
kylelaker:
```suggestion
def valid_branch_settings(self):
```

roles/common/files/uug_ansible_wrapper.py
kylelaker:
```suggestion
        if not validate_branch_settings(self):
```

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -388,7 +388,10 @@ def on_run_clicked(self, _):
         if not validate_branch():
             invalid_branch(self)
             return
-
+        
+        if not validate_config_settings(self):
+            continue
```
kylelaker: So this shouldn't actually be a `continue` statement. That's only valid in loops. Instead, can you replace this with a call to the logging library?
```suggestion
            logging.warn("Non-optimal user branch settings.")
```

roles/common/files/uug_ansible_wrapper.py
cfrome77: fixed with d4b7407

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -38,15 +39,15 @@
...
+    'git_url': DEFAULT_GIT_REMOTE,
     # All roles the user has ever chosen
     'roles_all_time': ["common"],
     # Roles to be used for this particular run
     'roles_this_run': ["common"],
 }
 NAME = "JMU CS VM Configuration"
 VERSION = "Spring 2019"
-
+DEFAULT_GIT_REMOTE = "https://github.com/jmunixusers/cs-vm-build"
```
kylelaker: This needs to be moved further up. Notice that it is used in Line 42 but not defined until Line 50? Can you move this either above the comment on line 26 or to line 40 before the  `USER_CONFIG = {`? As it is, the script is unable to run.

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -578,12 +574,101 @@ def get_distro_release_name():
...
+    
+    ls_remote = output.stdout.decode("utf-8")
+    
+    header = None
+    warning_prompt = None
+    
+    system_exists = system_version in ls_remote
+    chosen_exists = chosen_branch in ls_remote
+        
+    if chosen_remote == DEFAULT_GIT_REMOTE:
```
kylelaker: You can remove a whole indentation level by inverting the condition.

```python
if chosen_remote != DEFAULT_GIT_REMOTE:
    return True
```

And then unindent the rest of the if statements.

cfrome77:
I added an extra if check to account for your case. I realized I never had a check for when the chosen branch in the config exists on the remote but the systems Mint version does not exist.

kylelaker:
Thanks for taking care of this, @cfrome77!

Issue #189, opened 2018-08-29
Improve documentation and/or flexibility of the OEM role


kylelaker:
The OEM role is currently only really fit to be run on Linux Mint that is intended to be run in a virtual machine. There are many issues that come up if the OEM role is run on actual hardware (such as removing bluez libraries and failing because USB drivers are installed). These are necessary in the VM.

We should either:

 - have `oem-vm` and `oem-hw` roles that are intended for VMs or physical hardware respectively
 - document that OEM should only be run if the device is a virtual appliance and how to best bootstrap getting the `common` role on the machine initially.

ripleymj:
Just browsing the ansible setup variables, the following seem to be the best indicators of our VM:
        "ansible_bios_version": "VirtualBox", 
        "ansible_product_name": "VirtualBox", 
        "ansible_system_vendor": "innotek GmbH", 
        "ansible_virtualization_type": "virtualbox", 
        "ansible_virtualization_role": "guest",

kylelaker:
Oh wow. I really like `ansible_virtualization_role`. I'd like to see if that gets set consistently across VirtualBox, VM Ware, Parallels, and maybe a few others.

ripleymj:
## Bash script
| Task                | VM          | HW          |
| ------------------- | ----------- | ----------- |
| assert_mint         | Should      | Should      |
| apt update          | Must        | Must        |
| Install ansible/git | Must        | Must        |
| Run OEM playbook    | Must        | Must        |
| Zero empty space    | Must        | Should not  |

## OEM Playbook
| Task                | VM          | HW          |
| ------------------- | ----------- | ----------- |
| assert Mint         | Should      | Should      |
| apt update          | Must        | Must        |
| Remove unused       | Should      | Maybe not   |
| Remove unused deps  | Should      | Should      |
| Upgrade all         | Must        | Must        |
| Install deps        | Must        | Must        |
| Clean apt           | Must        | Maybe not   |
| Skel desktop        | Must        | Must        |
| Log directory       | Must        | Must        |
| Desktop README      | Must        | Must        |
| Hide mint welcome   | Should      | Should      |
| Create local share apps | Must    | Must        |
| Banish USB 2/3      | Must        | Must not    |

## Task shortcut Playbook
| Task                | VM          | HW          |
| ------------------- | ----------- | ----------- |
| VMTools dir         | Must        | Must        |
| Polkit wrapper      | Must        | Must        |
[...]

PR #225, opened 2018-11-24
Add VM checks in OEM role

kylelaker:
As discussed in #189, there are some things that should not be done in
when the `oem` role is run on a physical machine. This adds a check for
being a virtualization guest for certain tasks.

### DIFF
--- a/roles/oem/tasks/main.yml
+++ b/roles/oem/tasks/main.yml
@@ -1,5 +1,6 @@
 ---
 # tasks file for oem
+
 - name: Ensure this in Linux Mint
   assert:
       that:
@@ -23,13 +24,6 @@
   apt:
       name: "{{ packages_to_install }}"
       state: latest
-- name: Clean the apt database
-  command: apt-get clean
-  # This will always report as changed otherwise
-  changed_when: False
-  args:
-    warn: no
-  when: "ansible_virtualization_role == 'guest'"
 - name: Ensure skeleton Desktop directory exists
   file:
       path: /etc/skel/Desktop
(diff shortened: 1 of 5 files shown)


roles/oem/tasks/main.yml
```diff
@@ -66,3 +67,4 @@
   register: lspci_output
   failed_when: lspci_output.rc == 0
   changed_when: False
+  when: "ansbile_virtualization_role == 'guest'"
```
ripleymj: Plz to "ansible"

showalrm:
Does this not close #225? It seems like, looking at the list in that issue's discussion, that everything significant has been taken care of in some way.

kylelaker:
@showalrm That's a good point. If @ripleymj agrees this does everything he wanted in that issue, I'll edit the initial commit.

kylelaker:
I can't request changes on this PR since I opened it; however, the `dd` runs before the stuff to set up the config tool script. That probably wouldn't affect the goal of the `dd` too much.

If we don't care about this, go ahead and merge it.

If we do care, then we could probably either:

 - reverse the order of the roles in `oem.yml`
 - see if we can make the `dd` task a handler so it only ever runs once at the very end
    - can handlers have a `when`?
 - create a new role just for the `dd` and add it to the bottom of `oem.yml`.

kylelaker:
My preference in the above is probably to reverse the order in `oem.yml`. I think it's okay if `common` stuff is run on not-Mint, so we don't necessarily need to check that before installing the config tool script and Tux.

ripleymj:
Agreed, I think swapping them is better. Leaving it as is wouldn't hurt anything, as `common` does not remove anything to be zeroed. Moving `common` to be first in `oem.yml` makes this consistent with `local.yml`

ripleymj:
Adding @showalrm as an unbiased reviewer.

kylelaker:
I can't officially review, but FWIW, I approve.

kylelaker:
@ripleymj Hopefully 171ef24 takes care of that or at least gets us close.

roles/oem/tasks/main.yml
```diff
@@ -29,6 +29,7 @@
   changed_when: False
   args:
     warn: no
+  when: "ansible_virtualization_role == 'guest'"
 - name: Ensure skeleton Desktop directory exists
```
ripleymj: Do we need to keep this with it added to `common` now?

PR #226, opened 2018-11-25
Remove support for old versions of Java

### LINKED ISSUE #158: Figure out appropriate JVM version
Perhaps related to #154 

We're installing both default-jdk and openjdk-8. Default now depends on the openjdk-11 package, which identifies as Java 10.0.1. We probably should only have one JVM version installed.

kylelaker:
This removes Java 8 as well as DrJava which depends on it.

This must not be merged until we are certain which IDEs will be used in the spring.

This resolves #158.


ripleymj:
- [ ] chaoaj
- [ ] foxcj
- [ ] malloykp
- [ ] mayfiecs
- [ ] nortonml
- [ ] rileypb
- [ ] stewarmc
- [ ] weikleda

kylelaker:
As a note, we may not necessarily hear from everyone in @ripleymj's list above. I only explicitly requested responses for situations where this PR would cause an issue. We might consider waiting a few days just in case and then merging this. We can always revert it if needed.

kylelaker:
Confirmed that this does not install Java 11 and that jGRASP, Eclipse, and Logisim all still work.

PR #231, opened 2018-12-02
Move apt mirror setup from user to common

### LINKED ISSUE #227: Investigate moving mirror changes from user to common
The oem role does not currently use the JMU mirrors, and changing it might benefit the oem build process.

ripleymj:
Moving the mirror set up enables it to run during OEM setup

Closes #227


kylelaker:
Not sure if I just missed it, but don't we need to remove the lines to configure the mirror from `roles/user/tasks/main.yml`?

ripleymj:
Yeah, I think you're right. I missed a chunk in a hurry. I'll fix it up in a bit.

PR #236, opened 2018-12-06
Adjust logic for handling certain git branches

### LINKED ISSUE #235: config tool: Do not suggest branches that don't exist
If the user's settings are for the `master` branch while running `tessa`, they will get a dialog suggesting that they switch to the `tessa` branch. That isn't great advice because that branch doesn't exist. We should be more careful about only suggesting branches when they exist.

kylelaker:
This changes some of the warning messages, cleans up the conditions, and
allows for ignoring some warnings (we especially will find it annoying
if we always get prompt when we run against the master branch).

Closes #235


ripleymj:
I've been abstaining from the Python PRs, so merge this if you're ready.

PR #247, opened 2018-12-21
Ensure that a release exists as a branch

### LINKED ISSUE #246: config tool: Use better defaults on first launch
Currently, our `master` branch will build a Linux Mint 19.1 Tessa-based image when using `packer`. The repository does not yet have a `tessa` branch; however, when the JMU CS Configuration Tool is run for the first time, it will check `etc/os-release` to see what version of Mint is running and use that as the branch. While this behavior is generally correct, it is incorrect here as a `tessa` branch does not exist and therefore users will get a warning the first time they run the tool.

The defaults should be adjusted to select the current Mint version if it exists and to use `master` by default in all other cases. If a `tessa` branch were to be created, the user would get a warning suggesting they switch from `master` to `tessa` (this is a warning that the user can choose to ignore).

This should be completed before alpha images are built to ensure that our alpha testers don't get warning. We usually don't fork for the release until we ship beta images.

kylelaker:
If there is not a branch that is properly named, default to using the
master branch instead. It is the next-best option.

Closes #246

### DIFF
--- a/roles/common/files/uug_ansible_wrapper.py
+++ b/roles/common/files/uug_ansible_wrapper.py
@@ -79,8 +79,13 @@ def main():
 
     # The default value for the branch is the current distro release name.
     # Set this before parsing the configuration. If it can't be detected,
-    # it should get set to None
-    USER_CONFIG['git_branch'] = get_distro_release_name()
+    # it should get set to None. If the branch does not currently exist, it
+    # should be set to master.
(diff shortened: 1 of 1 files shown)


kylelaker:
I have tested this on a Tessa VM where the branch was set to `master` and on a Tara VM where the branch was set to `tara`.

Issue #257, opened 2019-01-17
Add 430 Install script


Advill:
As 430 requires a few specific programming languages, I feel it reasonable to add it's own install file.

The things that I know would be required right now are
* Ruby
* RVM (?)
* GHC
* GProlog

As well as the general text editing and similar programs installed alongside 361.

ripleymj:
Feel free to submit a draft PR if you want early feedback :)

Advill:
Haven't finished yet but I should have something done by the end of the weekend.

lam2mo:
I’m hoping we don’t need RVM honestly. IIRC @Advill ran into a non-Mint-specific issue that was resolved using RVM but I forget whether we determined it was necessary on Mint.

ripleymj:
Ok, I was going to be very surprised if 430 was using bleeding edge Ruby features.

kylelaker:
So Debian includes 2.3.3 ([source](https://packages.debian.org/stretch/ruby)) and Ubuntu and derivatives will have 2.5.1 ([source](https://packages.ubuntu.com/bionic/ruby)). Ruby 2.6.0 was *just* released not long ago ([source](https://www.ruby-lang.org/en/news/2018/12/25/ruby-2-6-0-released/)). I am okay with shipping 2.5.x for now. If they were shipping super old versions, I would lean more towards using `rvm`; however, on Ubuntu and Mint (which is what we plan to target with this tool), I think we're okay.

If you see an older version of Ruby getting installed, we can try to figure things out.

kylelaker:
Hopefully that addresses concerns about Ruby versions?

lam2mo:
Yes, it does for me. MacOS still ships with 2.3.7 by default so I'm actually fine with any of those. If students really want to be able to use features from the most recent version they can install it themselves. It's also possible that Dr. Simmons will choose to restrict them to using an older version anyway. I think as recently as last year we were telling folks not to use any functionality added after 2.0 (and linked them to the older documentation as a precaution).

ripleymj:
Probably the wrong place to leave this note to future self, but it seems relevant. If we know we have a bunch of people planning to use the VM this summer, let's try to get an updated build posted close to then so they don't have to patch so much on first boot.

lam2mo:
FWIW, "a bunch" here = ~12.

ripleymj:
What I hear you saying is THE LARGEST SUMMER INSTALLFEST IN UUG HISTORY!

[...]

PR #261, opened 2019-01-20
Add support for the Programming Languages class (CS 430)

Advill:
draft fix for #257

### DIFF
--- a/local.yml
+++ b/local.yml
@@ -14,3 +14,4 @@
     - { role: jgrasp, tags: ["cs149"] }
     - { role: finch, tags: ["cs101"] }
     - { role: y86, tags: ["cs261"] }
+    - { role: Programming-Languages, tags: ["cs430"] }
--- a/local.yml
+++ b/local.yml
@@ -14,4 +14,4 @@
     - { role: jgrasp, tags: ["cs149"] }
     - { role: finch, tags: ["cs101"] }
     - { role: y86, tags: ["cs261"] }
-    - { role: Programming-Languages, tags: ["cs430"] }
+    - { role: programming-languages, tags: ["cs430"] }
--- a/roles/common/files/uug_ansible_wrapper.py
+++ b/roles/common/files/uug_ansible_wrapper.py
@@ -37,6 +37,7 @@ COURSES = {
     'CS 261': 'cs261',
     'CS 354': 'cs354',
     'CS 361': 'cs361',
+    'CS 480': 'cs480'
 }
 USER_CONFIG_PATH = os.path.join(os.environ['HOME'], ".config", "vm_config")
 USER_CONFIG = {
--- a/roles/common/files/uug_ansible_wrapper.py
+++ b/roles/common/files/uug_ansible_wrapper.py
@@ -37,7 +37,7 @@ COURSES = {
     'CS 261': 'cs261',
     'CS 354': 'cs354',
     'CS 361': 'cs361',
-    'CS 480': 'cs480'
+    'CS 430': 'cs430'
 }
 USER_CONFIG_PATH = os.path.join(os.environ['HOME'], ".config", "vm_config")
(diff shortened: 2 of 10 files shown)


local.yml
```diff
@@ -14,3 +14,4 @@
     - { role: jgrasp, tags: ["cs149"] }
     - { role: finch, tags: ["cs101"] }
     - { role: y86, tags: ["cs261"] }
+    - { role: Programming-Languages, tags: ["cs430"] }
```
ripleymj: I think this needs to be lowercased to match the programming-languages role directory. I'm having trouble getting it to run past this point as is.

roles/programming-langs/vars/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+# vars file for programming-langs
+ruby_packages:
+  - ruby
+  - rvm
```
ripleymj: I don't see rvm in the Mint/Ubuntu repository on my VM

roles/programming-langs/vars/main.yml
```diff
@@ -0,0 +1,11 @@
...
+
+haskell_packages:
+  - haskell-platform
+
+prolog_packages:
+  - GProlog
```
ripleymj: This one seems to be "gprolog" here

roles/programming-langs/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+# tasks file for programming-langs
+
+- name: Install Ruby packages
+  apt: name={{ ruby_packages }} state=latest
+
+  # update ruby to latest stable version, as lam will generally use the same
+- name: Update Ruby with the command module
+  command: rvm get stable
```
ripleymj: I guess I misunderstood this when talking yesterday. I think rvm sets up a weird chain of dependencies if it's not in the apt sources. Line 5 above will install the Ubuntu Ruby version, then to run this, we'd have to pull RVM from outside the repositories. Once RVM's there, I'm assuming this would install a 2nd copy of Ruby, potentially at a higher version.

We should probably stick to either apt ruby or rvm ruby, but not both. My gut says apt ruby, unless we have a compelling reason to go outside the official repositories.

roles/common/files/uug_ansible_wrapper.py
```diff
@@ -37,6 +37,7 @@
     'CS 261': 'cs261',
     'CS 354': 'cs354',
     'CS 361': 'cs361',
+    'CS 480': 'cs480'
```
kylelaker:
```suggestion
    'CS 430': 'cs430'
```

ripleymj: I feel like my review was pointless, having missed this. :)

roles/programming-langs/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+# tasks file for programming-langs
+
+- name: Install Ruby packages
+  apt: name={{ ruby_packages }} state=latest
```
kylelaker: Unless @ripleymj disagrees, can we switch these to use the longer-form syntax? It'll make it easier to add new arguments in the future if necessary. I plan to open an issue to have us convert all the `apt` stuff to use the longer syntax so that it looks like every other task. It'd be nice to have this PR merged with the longer format.

```yaml
  apt:
    name: "{{ ruby_packages }}"
    state: "latest"
```

Advill:
Sorry about going dark for a bit...
All requested changes should be done, and I've switched to swi-prolog since the old one doesn't work anyways. Let me know if there is anything still to be fixed.

lam2mo:
Do we need more review on this or is it ready to merge? @ripleymj?

ripleymj:
Sorry, I just lost track of it.

Issue #258, opened 2019-01-17
Add process for adding new classes to CONTRIBUTING.md


kylelaker:
This is a fairly involved process at this point and I'd like to see it documented. This can probably done along with #257 since we're adding another class.

kylelaker:
Closed by #262

PR #262, opened 2019-01-20
Add process for adding a new class to CONTRIBUTING.md

Advill:
draft fix for #258

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -143,3 +143,8 @@ install that can be done with a `command` module call instead of a more specific
 module generally should, as using a seperate module would require installing
 that module on every VM, as we are not using ansible as it was designed.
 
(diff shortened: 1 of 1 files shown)


ripleymj:
Depending on how #261 goes, we might need to untangle the git history of this branch before merging.

Advill:
Oh yeah you're right, thought I was safe using branches, turns out I was mistaken.

kylelaker:
If you want to, you should be able to `git rebase -i master` and then just not pick the 430-related commits. Alternatively, we can just clean this up after #261 gets merged.

ripleymj:
Branches are still safe, but it looks like you got a branch of a branch, rather than a 2nd branch off master. Do the thing Kyle said, then force push it back here.

Advill:
Ok, This should be untangled from the programming-languages update, check me to be sure though.

PR #264, opened 2019-01-22
Add checkstyle plugin installation to Eclipse

ripleymj:
Please give this a try when you get a moment. I think the `creates` argument prevents re-running the install, but I left it outside the usual installation block so people could run the role and add checkstyle to an existing install. If possible, try this on a machine with checkstyle installed to `$HOME/.eclipse` via the usual methods, then see if Eclipse loses its mind with two installs present.

I'll do a backport if so encouraged.


kylelaker:
I do not use Eclipse and would not be able to suitably test this. The actual config does look good to me.

kylelaker:
Please do backport this.

PR #274, opened 2019-04-08
Clean up Packer options to reflect new functionality. Set default audio

### LINKED ISSUE #249: packer: No audio in exported VMs
When creating an appliance using Packer, audio does not work in the resulting VM. The audio settings are disabled in VirtualBox and it doesn't seem we have any existing commands to enable them. We should try to get this working before shipping the VM in the spring if we can.

ripleymj:
Since https://github.com/hashicorp/packer/pull/7017, we have not had working audio, regardless of VirtualBox version. This PR defaults us to PulseAudio on the host, but can be overridden by command line flag when building on Windows (daudio) or Mac (core). I'm certainly open to better approaches if they develop in Packer or Virtualbox, but I think this is the expedient fix for now.

Also clean up a few additional Packer options based on the new features available.

Closes #249

### DIFF
--- a/packer/oem-build.json
+++ b/packer/oem-build.json
@@ -10,6 +10,7 @@
     "git_branch": "master",
 
     "headless": "true",
+    "audio": "pulse",
 
     "output_dir": "{{pwd}}/artifacts",
     "mirror_url": "http://mirror.cs.jmu.edu/pub/linuxmint/images/stable",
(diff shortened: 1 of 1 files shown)


ripleymj:
An image built with this configuration is available for download: https://w3.cs.jmu.edu/uug/

packer/oem-build.json
```diff
@@ -23,12 +24,14 @@
       "type": "virtualbox-iso",
       "headless": "{{user `headless`}}",
       "http_directory": "http",
+      "vm_name": "{{user `vm_name`}} {{user `semester`}}",
+
+      "cpus": 2,
+      "memory": 4096,
```
kylelaker: Are we confident that this jump will be fine on the majority of machines? If we still have people coming in with only 4GB of RAM, this may be problematic. Not a rejection, but worth discussing

packer/oem-build.json
```diff
@@ -10,6 +10,7 @@
     "git_branch": "master",
 
     "headless": "true",
+    "audio": "pulse",
```
kylelaker: I am curious if we should just change the build-on-Linux and build-on-Windows scripts to just run packer and specify a value for audio. Those scripts are not used at all (as far as I am aware).

We could then just use the `null` audio driver here to ensure that the packer build never fails (even if audio doesn't work in the resulting VM). The relevant script would pass the right argument.

packer/oem-build.json
ripleymj: I've been resisting this change for a while, but I think we've reached a point where a majority of people will benefit from this change, and if we do encounter a machine so under-powered, RAM is an easy enough GUI tweak after import. It's also more the 8GB machines that scare me here. I think 4GB is going to have a bad experience regardless, but at 8, we're taking 50% of the machine.

kylelaker: Works for me!

packer/oem-build.json
ripleymj: I'm not excited about introducing a layer of indirection for something so minor. The scripts should probably be archived at this point, as the output has diverged so much from Packer. As far as I know, I'm the only one who's ever built this on Windows, and I can live with a command line flag. The script would also introduce another layer when trying to override other Packer options.

kylelaker: Sounds good

PR #277, opened 2019-04-09
Mega-backport for summer users

ripleymj:
This is an attempt to backport most changes to software that we've accumulated this semester. As we have potential summer users of the tessa branch, I think we need to make more of an attempt than usual to keep it up to date. I'm certainly open to discussion about its contents.


ripleymj:
Ugh, wrong base branch

PR #289, opened 2019-07-01
Update comment in default apt sources list file

kylelaker:
The current file tells users to not make modifications in that
particular file; however, we do nothing to tell users what the correct
thing to do would be, especially since some guides for Linux Mint
specifically suggest putting things in official-package-repositories.


kylelaker:
Merging with the two approvals.

PR #291, opened 2019-07-02
Allow apt update to fail

ripleymj:
Ansible does not allow any mirror to fail when performing an update.

Same as #290


kylelaker:
We should look into causes or alternatives at a later time. Approving for now.

PR #295, opened 2019-07-08
Change graphics controller to VMSVGA

### LINKED ISSUE #280: Consider new VBox 6 graphics driver
We should investigate these new settings, profile their performance, and possibly add steps to make sure modules will compile:

https://blogs.oracle.com/scoter/oracle-vm-virtualbox-6-3d-acceleration-for-ubuntu-1804-guest

ripleymj:
The driver seems to work better in the casual testing I've done. Let's switch as part of the fall release testing.

Closes #280

### DIFF
--- a/packer/oem-build.json
+++ b/packer/oem-build.json
@@ -32,6 +32,7 @@
 
       "vboxmanage": [
         ["modifyvm", "{{.Name}}", "--clipboard", "bidirectional" ],
+        ["modifyvm", "{{.Name}}", "--graphicscontroller", "vmsvga" ],
         ["modifyvm", "{{.Name}}", "--mouse", "usbtablet"],
         ["modifyvm", "{{.Name}}", "--pae", "on"],
         ["modifyvm", "{{.Name}}", "--rtcuseutc", "on"],


kylelaker:
The blog post from the issue seemed to mention some manual intervention in the guest. Did I misunderstand or is it no longer necessary? Does this introduce a hard dependency on VBox 6?

ripleymj:
I'm not seeing anything, just setting the VM preference, making sure the guest additions are installed, and a command to validate that the driver is active. We can certainly run that to validate for our own testing, but I don't see it as a requirement for the end-user.

I believe this would raise the minimum level to 6, but we're already pushing the 6.x drivers and building the OVA on 6. VBox 6 will have been out around a year by the time we ship Tina, so I think I'm ok with setting it as the baseline.

ripleymj:
VMSVGA seems to work better all of the time. The desktop doesn't tear as much when resizing, the screensaver doesn't get stuck half-engaged, and the Mint logo doesn't appear yellow.

Advill:
Do I need to build the vm myself as far as testing this?

ripleymj:
I've got a test 19.2 build you can use if you want: https://linode2.ripleymj.net/~ripleymj/artifacts/ or just toggle the VMSVGA driver in the VirtualBox settings on any existing copy of 19.1.

PR #297, opened 2019-07-08
Fix or squelch ansible-lint warnings

ripleymj:
Keeping a clean ansible-lint report for local.yml and oem.yml would be good. I'm slowly working down the list.

Currently ignoring:
[403] Package installs should not use latest


kylelaker:
Clearly it'd be cool to start adding real CI/CD to this project in some way, but temporarily, should we look into adding a Travis build that just runs `ansible-lint`?

ripleymj:
It's somewhere on my list of considerations, but we'd need to do something clever since Travis only has ancient Ubuntu. I figure getting a clean report is probably the first step here.

I also just tried a generic yaml linter, which might be a good addition, but it seems to have very different opinions about style vs ansible-lint.

ripleymj:
And before anyone runs off doing this, we should probably have some evaluation of the CI services, and see if Travis is even a decent fit.

ripleymj:
This PR eliminates all ansible-lint warnings now, with exceptions for the following rules:
301: Commands should not change things if nothing needs doing
305: Use shell only when shell functionality is required
306: Shells that use pipes should set the pipefail option
403: Package installs should not use latest

403 is excluded globally, as we intentionally do that all over the place. The rest are targeted with the "noqa" flag. 306 should be reviewed later, but modifying that right now scares me.

I also rolled a fair number of yamllint cleanups into this, but that's nowhere near silent yet, and I'm not sure it easily will be.

roles/oem/tasks/vm_only_post.yml
```diff
@@ -15,7 +15,7 @@
 # The USB 2/3 controllers are only available with the Virtualbox
 # Extensions which are non-free software and available only under the Oracle
 # PUEL. We should ensure we have not accidentally included them.
-- name: Validate missing USB 2/3 controllers
+- name: Validate missing USB 2/3 controllers # noqa 306
```
kylelaker: Can we document why we don't use pipefail (or at least say that we plan to re-evaluate later)?

roles/user/handlers/main.yml
```diff
@@ -6,15 +6,15 @@
...
+    "Changes have been made to your machine that
+    will only take effect when you log out or
     when you reboot. Please do this immediately."
   environment:
     XDG_RUNTIME_DIR: "/run/user/{{ item.uid }}"
   become: yes
   become_user: "{{ item.user }}"
   loop: "{{ real_users }}"
-- name: Restart update manager
+- name: Restart update manager # noqa 305
```
kylelaker: Does this actually need shell or can we use command? If not, can we document why?

PR #298, opened 2019-07-21
Mirror transport

### LINKED ISSUE #292: Consider alternative approaches to apt update failure
After #290 and #291 taught us that Ansible will consider failure of any mirror as fatal, we made the step optional. Adding both the CS mirror and the Ubuntu upstream was intentional, to increase odds of success, but with this failure early in the playbook, software installation cannot succeed.

Opening this issue for discussion on alternate approaches that would allow apt source failover.

ripleymj:
Here's an attempt at #292. Apt has a "mirror transport" instead of http/https. In brief testing, it appears to allow more explicit configuration of mirror priority, and Ansible doesn't trip over itself when a mirror is unavailable.

If merged, closes #292

### DIFF
--- a/roles/common/files/mint.mirror
+++ b/roles/common/files/mint.mirror
@@ -1,0 +1,2 @@
+https://mirror.cs.jmu.edu/pub/linuxmint/packages	priority:1
+http://packages.linuxmint.com	priority:2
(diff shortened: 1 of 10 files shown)


ripleymj:
This breaks the Mint Update Manager, so it may not be usable.

VirtualBox_Mint Tina_21_07_2019_20_09_23

ripleymj:
This also seems to prevent Ansible from finishing the Packer build, so I'm inclined to scrap the idea at this point.

PR #300, opened 2019-07-21
Remove i386 architecture from dpkg

### LINKED ISSUE #294: Remove i386 support
Mint seems to autoconfigure support for i386 packages, and pulls the package catalogs for both architectures on every update. Since we have no i386 packages, let's remove this during OEM setup.

ripleymj:
Closes #294

### DIFF
--- a/roles/oem/tasks/vm_only_pre.yml
+++ b/roles/oem/tasks/vm_only_pre.yml
@@ -3,3 +3,8 @@
 # Tasks for the oem role to run at the beginning of playbook execution
 - name: Remove unneeded stock packages
   apt: name={{ packages_to_remove }} state=absent purge=yes
+- name: Remove i386 architecture
+  lineinfile:
+    path: /var/lib/dpkg/arch
+    state: absent
+    line: 'i386'


kylelaker:
We perform an upgrade and an install immediately after the i386 architecture is removed. Should the package cache be updated before we upgrade and after we remove the architecture?

Is there a reason to not just remove the architecture for real hardware as well? Are we are sure we even would work on i386? Do we want to accept that we wont't after Mint 20 anyway?

ripleymj:
If we weren't running in the tightly controlled OEM context, I'd say possibly another package cache update, though I feel like we already have too many of them. I'm only pushing this because Mint ships without any i386 packages installed, so the upgrade shouldn't trigger any i386 package installations. Shortly after that, we erase the entire package database.

I didn't want to do this for real hardware, as we're not as tightly constrained on space, and I think the probability of someone doing something weird like installing a 3rd party driver bundle with some 32 bit dependency, or installing Steam is higher. If Mint 20 removes support altogether, then we can drop this.

PR #303, opened 2019-07-23
Update ROS apt-key to match new GPG key

### LINKED ISSUE #302: ROS Melodic install error
The following error occurred on my latest attempt to update with cs345 checked (image-sp19e):
`019-07-22 17:21:04,540 p=16549 u=root |  TASK [robot-pkgs : Install robot development packages] 2019-07-22 17:21:05,361 p=16549 u=root |  fatal: [localhost]: FAILED! => {"changed": false, "msg": "No package matching 'ros-melodic-desktop-full' is available"}`

This is strange because the package reference does exist in the Packages file ([http://mirror.umd.edu/packages.ros.org/ros/ubuntu/dists/bionic/main/binary-amd64/Packages](http://mirror.umd.edu/packages.ros.org/ros/ubuntu/dists/bionic/main/binary-amd64/Packages)) on the mirror

cfrome77:
Updated the key, since the ROS folks recently released a new GPG key that deprecates the old one.
Link documenting the change: [https://discourse.ros.org/t/new-gpg-keys-deployed-for-packages-ros-org/9454](https://discourse.ros.org/t/new-gpg-keys-deployed-for-packages-ros-org/9454)

closes #302

### DIFF
--- a/roles/robot-pkgs/tasks/main.yml
+++ b/roles/robot-pkgs/tasks/main.yml
@@ -3,12 +3,12 @@
 - name: Add ROS key
   apt_key:
     keyserver: ha.pool.sks-keyservers.net
-    id: 421C365BD9FF1F717815A3895523BAEEB01FA116
+    id: C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654
     state: present
-- name: Add ROS repository
+- name: Add ROS repositories
   template:
     src: ros-packages.j2
-    dest: /etc/apt/sources.list.d/ros-packages.list
+    dest: /etc/apt/sources.list.d/ros-package-repositories.list
     owner: root
     group: root
     mode: 0644


cfrome77:
I was unable to rebase the commit to update the username and email to match my account. I kept getting `remote: error: You're not authorized to push to this branch`

ripleymj:
It looks like you may have pushed a branch to the jmunixusers repo, instead of the more traditional personal fork. Any branch on jmunixusers falls under the PR review rules, which is why it won't let you push, but we've never found a way to prevent branch creation.

The key server infrastructure has been having severe issues lately with people spamming/DoSing well-known keys. I forgot we were using it to import this, but I think there are recommendations on alternate safer ways to import a key now.

Go ahead and leave this PR for now, but if we go another route on this, I think it'd be best to close this PR, clean up the branch, and then have you reopen from a personal fork. I think we're also on the edge of deciding whether CS354 is worth continuing here, as we have no way to test it against actual robots, and no confirmed users. I'm hesitant to ask you to spend any more time on this if it's just getting shut down :(

ripleymj:
I swear there was a write-up somewhere about using an out-of-band method to import keys, but LWN specifically calls out the SKS server for being a mess: https://lwn.net/Articles/792366/

The ROS community recommends using hkp://keyserver.ubuntu.com:80 in the article you linked, which might be a slightly better stop-gap while we figure things out.

cfrome77:
OK, sounds good. I totally didn't realize that I wasn't using my forked copy. I'll wait to here what the consensus is on continuing with the upkeep of the CS354 role.

kylelaker:
Based on the `apt_key` [module docs](https://docs.ansible.com/ansible/latest/modules/apt_key_module.html), we should probably use the `url` field to specify http://packages.ros.org/ros.asc which will allow us to avoid using the SKS keyserver pool. For security reasons, we should continue specifying the id as `C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654` in case the packages server is compromised and the remote key file changes unexpectedly. 

Since this is technically an issue in a currently-supported release, I also think the final patch should eventually get backported to `tessa` (even though we're prepping for Tina). But let's work on getting something into master first.

@ripleymj Maybe an issue should be opened for discussing the deprecation of the CS 354 tag?

roles/robot-pkgs/tasks/main.yml
```diff
@@ -3,12 +3,12 @@
...
     keyserver: ha.pool.sks-keyservers.net
-    id: 421C365BD9FF1F717815A3895523BAEEB01FA116
+    id: C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654
     state: present
-- name: Add ROS repository
+- name: Add ROS repositories
   template:
     src: ros-packages.j2
-    dest: /etc/apt/sources.list.d/ros-packages.list
+    dest: /etc/apt/sources.list.d/ros-package-repositories.list
```
kylelaker: This change will actually result in an issue for existing users. Rather than replacing the file, it will create a new one. This will result in two identical files with different names. Even though the current name isn't ideal, the extra effort to fix it isn't really worth it.

But thank you for pointing it out! Perhaps if we phase out 354, we can make sure to fix this if it ever gets phased back in.

cfrome77: OK, thanks for pointing that out. I will be sure to not include this change in my forked version, If there are not objections, I will go ahead and open new pull request from my forked repo, to fix this issue for current users.

ripleymj:
@cfrome77 I think we've got what we need here, so close this PR when you're ready. If you can work up a new branch with:

- file name changed
- author/email set
- HTTP key fetch

I'm still debating taking CS354 out of the GUI tool until we can figure out how to support this long term, but for now, let's try to keep the role in a working state.

cfrome77:
opened #304 to reflect changes discussed here

PR #317, opened 2019-09-01
Add VSCode support for CS149

### LINKED ISSUE #315: Need to install VSCode for CS149
VSCode, Java Extension Pack, and Checkstyle for Java need to be installed for Dr. Bowers.

ripleymj:
Adds Microsoft key and repo, installs Code, and adds Java/Checkstyle plugins per user. Like Eclipse, it leaves a checkstyle.xml configuration to the user.

Closes #315

### DIFF
--- a/roles/vscode/tasks/main.yml
+++ b/roles/vscode/tasks/main.yml
@@ -1,0 +1,30 @@
+---
+# tasks file for vscode
+- name: Add Microsoft key
+  apt_key:
+    id: BC528686B50D79E339D3721CEB3E94ADBE1229CF
+    url: https://packages.microsoft.com/keys/microsoft.asc
+    state: present
+- name: Install VSCode Repository file
+  apt_repository:
+    repo: deb https://packages.microsoft.com/repos/vscode stable main
+    state: present
+    filename: vscode
+- name: Install VSCode
+  apt:
+    name: 'code'
+    state: latest
+- name: Install Java pack plugin
+  command: /usr/bin/code --install-extension vscjava.vscode-java-pack
+  args:
+    creates: '{{ item.homedir }}/.vscode/extensions/vscjava.vscode-java-pack*'
+  become: yes
+  become_user: "{{ item.user }}"
+  loop: "{{ real_users }}"
+- name: Install checkstyle plugin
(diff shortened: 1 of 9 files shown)


roles/vscode/files/vscode.list
```diff
@@ -0,0 +1 @@
+deb [arch=amd64] https://packages.microsoft.com/repos/vscode stable main
```
kylelaker: I know they recommend specifying the architecture, but is it really necessary if amd64 is the only one we support anyway?

roles/vscode/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
+---
+# tasks file for vscode
+- name: Add Microsoft key
+  apt_key:
+    id: BC528686B50D79E339D3721CEB3E94ADBE1229CF
+    url: https://packages.microsoft.com/keys/microsoft.asc
+    state: present
+- name: Install VSCode Repository file
```
kylelaker: Is there a particular preference for this implementation over the following?

```
- apt_repository:
    repo: deb https://packages.microsoft.com/repos/vscode stable main
    state: present
    filename: vscode
```

kylelaker: FWIW, an `apt update` is automatically performed when an `apt_repository` entry changes by default.

roles/vscode/files/vscode.list
ripleymj: Only because Microsoft provides it that way.

roles/vscode/tasks/main.yml
ripleymj: Stole this from robots, but there we're using a template to set the Ubuntu version, so I guess it's less necessary here. And I didn't realize apt-repository was a thing.

kylelaker: Yeah, I am a little curious if we should evaluate our use of .list files. Think in most the cases we could use the variable expansion except maybe common where we explicitly want to override entire files... But resolving this in other places can be a separate issue at a later time.

roles/vscode/files/vscode.list
kylelaker: I think this file can be removed now

ripleymj: Derp

PR #318, opened 2019-09-01
Force all pending dpkg operations to complete

### LINKED ISSUE #314: dpkg gets stuck
Need to run `dpkg --configure -a` manually.

ripleymj:
This should force packages that are half-installed to be finished before we attempt anything else. If there's nothing pending, this should be a no-op. Thought about querying for stuck packages and only doing this if absolutely necessary, but decided it wasn't worth the effort, since it will silently proceed if everything if fine.

Closes #314


kylelaker:
I am concerned that this will either always report as changed. Would it be possible to do something like:

```
register: dpkg_command
changed_when: dpkg_command.stdout != ""
```

PR #319, opened 2019-09-01
Handle Enter/Return press

### LINKED ISSUE #248: Handle <Tab>/<Return> in Settings dialog
Currently hitting tab and return in the settings dialog may not do what we want, especially return. This should be adjusted for both of those.

kylelaker:
Closes #248


kylelaker:
Checked and it looks like tab is already handled like you'd expect. So pressing enter/return just clicks 'OK' for you.

PR #320, opened 2019-09-01
Fix USB validation

### LINKED ISSUE #307: Investigate pipefail options, and explicit shell choice
Follow-on from #297, find out what pipefail would take to implement here:

https://github.com/jmunixusers/cs-vm-build/blob/a8acb6d927b37b133835248a60ef8eb374deb160/roles/oem/tasks/vm_only_post.yml#L18

kylelaker:
We can just check `stdout` from the command itself. This way shells and pipes don't matter.

Resolves #307


kylelaker:
Meant to make this a draft but seems like I don't get to change that after the fact...

kylelaker:
Okay, testing complete. I have seen a failure with USB 3 enabled and succeeds with USB1.1 selected.

PR #324, opened 2019-09-03
Put local copy of repository in /tmp

### LINKED ISSUE #323: ansible-pull does not clean up if interrupted
Please add --clean to wrapper script launch and README, or set ansible to pull the git repo to tmpfs so it is erased on reboot.

showalrm:
After a restart, the local copy of the playbook repository will be erased.

We may wish to add --clean as well, so I'm open to suggestions.

Resolves #323

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ ansible-playbook -i hosts -c local -K -t TAGS local.yml
 or directly from GitHub:
 
 ```
-ansible-pull -U https://github.com/jmunixusers/cs-vm-build --purge -i hosts -K -t TAGS
+ansible-pull -U https://github.com/jmunixusers/cs-vm-build --directory /tmp --purge -i hosts -K -t TAGS
 ```
 where TAGS is a comma separated list (with no spaces) of
 cs101, cs149, cs159, cs261, cs354, and/or cs361 as appropriate.
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ ansible-playbook -i hosts -c local -K -t TAGS local.yml
 or directly from GitHub:
 
 ```
-ansible-pull -U https://github.com/jmunixusers/cs-vm-build --directory /tmp --purge -i hosts -K -t TAGS
+ansible-pull -U https://github.com/jmunixusers/cs-vm-build --directory /tmp/cs-vm-build --purge -i hosts -K -t TAGS
 ```
 where TAGS is a comma separated list (with no spaces) of
 cs101, cs149, cs159, cs261, cs354, and/or cs361 as appropriate.
--- a/roles/common/files/uug_ansible_wrapper.py
+++ b/roles/common/files/uug_ansible_wrapper.py
@@ -426,6 +426,8 @@ class AnsibleWrapperWindow(Gtk.Window):
             USER_CONFIG['git_url'],
             '--checkout',
             USER_CONFIG['git_branch'],
+            '--directory',
+            '/tmp',
             '--purge',
             '--inventory',
             'hosts',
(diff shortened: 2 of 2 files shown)


ripleymj:
On my copy of the VM, `/tmp` appears to be on the root FS, so it wouldn't be wiped on reboot. I need to see if there's some namespacing trickery in play though.

kylelaker:
I believe, and simple experimentation upholds, that even though `/tmp` is not a `tmpfs`, it still gets cleared after a reboot. I don't think that Mint will clear files from `/tmp` after a certain time frame but I don't think that's an issue here.

kylelaker:
So accepting the following:
 - my previous suggestion probably won't get much traction
 - we need to get #325 merged

Let's talk about getting this merged as-is.

`--directory` cannot be set to `/tmp` because `/tmp` already exists and therefore `git` cannot checkout there. The directory would need to be set to `/tmp/cs-vm-build` or something like that. I'd like to see the Python script actually generate a short (6-character?) random string and append it to the end like: `/tmp/cs-vm-build-<random-string>` so that even if the directory weren't removed, the next run wouldn't use the same one. The README can recommend the other option.

The error message I get with the current implementation is:

```
fatal: destination path '/tmp' already exists and is not an empty directory.
```

The random string function could just look something like:

```python

def random_string(length=6):
    """
    Generate a string with random characters of a given length.
    """
    characters = string.ascii_uppercase + string.digits
    return "".join(random.choices(characters, k=length))
```

But at least changing to `/tmp/cs-vm-build` will be necessary.

kylelaker:
I am happy to add the changes I mentioned to `use-tmpfs` if you don't have the time to work on this right now :)

Crosse:
I’d caution against reinventing the wheel here and just recommend using `tempfile.mkdtemp()`. You can either clone directly into the resulting directory (since it’ll be empty) or clone into a subdirectory.

Advill:
Just reading over this now while looking at backport stuff. Can I get a better description of what problem exists here and how we're solving it? I agree a temp directory that is not wiped on startup would be helpful for debugging but I'm not sure what ansible is currently doing that is a problem.

ripleymj:
Excellent question. In working on #318, I tried killing Ansible mid-run to force dpkg errors, and in doing so, left behind a stale copy of the repo that it had pulled. We run with --purge, so that on completion, it deletes itself, but had never considered an interrupted run. I thought at the time, that adding --clean or a tmp directory would be a nice feature to prevent a potential problem, but like I mentioned on #325, this has never been seen in the wild.

kylelaker:
@Crosse: Oh yeah! That'd definitely be better. I didn't even consider that the issue was that the directory was __non-empty__ not the pure existence of the directory. Thanks for the suggestion. @showalrm please do that and ignore my suggestion.

---

@Advill: I believe the issue is that if Ansible and is interrupted (SIGINT or something like that), `/root/.ansible` doesn't get cleaned up/removed correctly. It only gets removed at the end of a _completed_ run. When the directory exists, `ansible-pull` refuses to run because it cannot perform a `git clone` or `git pull` in that directory correctly. It'd be cool to automatically take care of that.

As @ripleymj reminded us in #325, we haven't seen this in the wild. It _could_ be caused if a student were to close the wrapper during the Ansible run which I believe implicitly sends SIGINT to the called process, and therefore Ansible would not clean up correctly. But the Quit button is disabled while Ansible is running and `^C` doesn't work in the VTY they have in the GUI, so they'd have to use the `X` button on the window.

showalrm:
Hopefully now everything looks good! Let me know if there's anything else that needs to be done!

showalrm:
@kylelaker Thanks for your help with this!

PR #325, opened 2019-09-09
Tina megabackport 1

ripleymj:
With the InstallFest on Wednesday, let's backport the best of the fixes since the 101 sessions.


ripleymj:
I'm less convinced after seeing the conversation in #324 that we should rush this through tonight and backport it. We've never actually seen the problem we're proposing to fix in the wild, and to be effective, it will require a rebuild on the OS image.

PR #334, opened 2019-09-15
Cleanup inventory file

### LINKED ISSUE #327: README manual steps do not work for Ubuntu 19.04
Current manual steps only recommend `apt-get install ansible git`, but Ansible fails to launch on Ubuntu 19.04 because `/usr/bin/python` is missing. We should either recommend installing the `python` (2.7) package, or alternate steps that specify a working interpreter.

Also, Ansible immediately warns that `aptitude` is missing, and that it's falling back to `apt-get`, so we might want to recommend installing `aptitude` as well.

kylelaker:
Adds support for python3 and removes the need to specify '-c local' on
the CLI.

Resolves #327

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ using these tasks and the VM is available
 apt-get install ansible git
 git clone https://github.com/jmunixusers/cs-vm-build
 cd cs-vm-build
-ansible-playbook -i hosts -c local -K -t TAGS local.yml
+ansible-playbook -i hosts -K -t TAGS local.yml
 ```
 or directly from GitHub:
 
--- a/hosts
+++ b/hosts
@@ -1,1 +1,1 @@
-localhost
+localhost ansible_connection=local ansible_python_interpreter=/usr/bin/python3
--- a/roles/user/vars/main.yml
+++ b/roles/user/vars/main.yml
@@ -6,7 +6,7 @@ global_profile_path: /opt/csvmprofile
 user_dependencies:
   - default-jdk
   - libnotify-bin
-  - python-psutil
+  - python3-psutil
   - unzip
   - vim
   - vim-gnome


kylelaker:
@ripleymj @Advill: Is there anything else we need to do before merging this?

ripleymj:
I was kidding myself into believing that I was going to research why we were getting Python3 GUI wrapper, Python2 ansible master, and Python3 ansible remote, but honestly at this point, I'm not sure I'll get to it any time soon.

kylelaker:
So [PEP 394](https://www.python.org/dev/peps/pep-0394) specifies how `/usr/bin/python` should work and what version of Python it should point to. When Mint 19 came out, it recommended that it should be a symlink to `/usr/bin/python2`. Now it has different recommendations. The PEP also recommended that developers should specify a particular version in their `#!` if their script only worked with one version or just use `/usr/bin/python` if their script worked with both. Ubuntu seems to now just not be providing a `/usr/bin/python` unless you explicitly install the `python2` package. The correctness of this isn't really for us to decide, but just work around (or open a bug report upstream).

Ansible as of 2.5 support both python2 and python3. Luckily 2.5 is what we have installed on the current versions of Mint. The GUI script in this repo supports only python3 (and therefore that's what the wrapper does) and will fail if run with python2. `ansible-pull` works with either python2 or python3. Its `#!` on Mint is `/usr/bin/python` and on the latest Ubuntu is `/usr/bin/python3`. This is why even though Ubuntu doesn't have python or python2 by default, that still works (and why it continues working on Mint).

`ansible-pull` does some magic to launch `ansible-playbook` because it respects your configuration in `ansible.cfg` in the repo you pull. Since you can specify the interpreter, it will try to launch `ansible-playbook` with that and then it will fall back to `/usr/bin/python` if it isn't specified (again, maybe this should be a bug report against Ubuntu's `ansible` package to patch the default). `ansible-playbook` will, as of this PR be launched with python3. Since `ansible-playbook` will use the interpreter, all of its children will be launched with python3 as well.

Could we launch the wrapper with python3 on Mint? Probably. Is it worth it? Probably not. It will be launched with python3 by default in a future version of Mint and the current version of Ubuntu.

So on Mint 19.2, our process tree will look something like this:

    \_ python3 wrapper
    |  \_ pkexec
    |    \_ python2 ansible-pull
    |      \_  python3 ansible-playbook
    |        \_ python3 task1
    |        \_ python3 task2


And on Ubuntu Disco, our process tree will look something like this:

    \_ python3 wrapper
    |  \_ pkexec
    |    \_ python3 ansible-pull
    |      \_  python3 ansible-playbook
    |        \_ python3 task1
    |        \_ python3 task2

PR #341, opened 2019-10-02
Update eclipse download link

nuttercd:


ripleymj:
Hold on, somehow you're 112 commits behind master, which is probably a fork from your last PR. Please let me work with you on this IRL, so we can get a clean PR.

PR #345, opened 2019-10-16
Remove IDE controller during VM creation

### LINKED ISSUE #309: Remove IDE controller from Packer config
At one point, Packer required the IDE controller for VM Additions installation, but I think it can be removed now. SATA should be sufficient for us.

ripleymj:
Turns out, the IDE controller wasn't needed. We already have a SATA CD, and the initial additions installation is done by uploading the iso over sftp.

Closes #309


kylelaker:
Do you happen to have a .iso built with this? Asking hopefully, but can make one if one isn't ready. Want to test that the 'Devices > Insert Guest Additions CD image...' and 'Devices > Optical Drives' menu options work as expected without the IDE controller, just in case a user wants to attach a CD or update guest additions

ripleymj:
I built an image but didn't publish it. All the CD features appear to work fine without IDE.

PR #350, opened 2019-12-05
Add custom mintreport script

kylelaker:
This will present a "Detected problem" that can't be ignored if the tool hasn't been run. It's in `common` so it gets included in OEM and so we can add more features in the future (like detecting if there is a newer version available in `git` than what the user last ran). Currently it just checks to see if `/opt/csvmprofile` exists, which is a file created only in `user`, which is a way to detect whether the tool has ever been run.

This will only really have an effect on Mint 19.3 or newer; however, it should not cause any issues on older releases either.

### DIFF
--- a/roles/common/files/CustomMintReportInfo.py
+++ b/roles/common/files/CustomMintReportInfo.py
@@ -1,0 +1,42 @@
+import os
+import gettext
+import gi
+import subprocess
+
+gi.require_version('Gtk', '3.0')
+from gi.repository import Gtk
+
+from mintreport import InfoReport, InfoReportAction
+
+class Report(InfoReport):
+
+    def __init__(self):
+
+        self.title = "Run JMU CS Config Tool"
+        self.icon = "edu.jmu.uug.tux"
+        self.has_ignore_button = False
+
+    def is_pertinent(self):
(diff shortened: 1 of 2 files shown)


kylelaker:
Not really sure the best way to handle that failure. Do we ignore the lint because we never know whether something changed? Or do we just say nothing ever changes because it's a fairly inconsequential change that would happen on a reboot or login/logout anyway?

ripleymj:
The only thing I'm absolutely convinced of is that we need to keep the lint tests passing overall, or we'll quickly lose the value of them. If the cache update can report change status based on exit code or something, that's probably ideal. For other plays, we've just marked them as exempt from the linter with "  # noqa 301" (space space hash space - to keep yamllint happy)

kylelaker:
Definitely no disagreement from me there. Just not sure 

```
  # noqa 301
- name: Update icon cache
  command: /usr/sbin/update-icon-caches /usr/share/icons/*
```

or 

```
- name: Update icon cache
  command: /usr/sbin/update-icon-caches /usr/share/icons/*
  changed_when: False
```

or trying to figure out whether the cache updated (but there doesn't seem to be any sort of return value info or documentation on where the cache is and I don't want to step through with `strace` to find what files are being touched) is preferred.

PR #352, opened 2019-12-08
Add Packer lint and clarify YAML linting

ripleymj:
Been thinking about the Actions testing a bit more. I think this cleans up a few things from the GitHub beta around Python installation, and then splits out "lint.yml" into more granular tests in case we find more things that can be tested.


.github/workflows/yamllint.yml
```diff
@@ -0,0 +1,30 @@
+---
+name: YAML lint workflow
+on:
+  pull_request:
+    paths:
```
kylelaker: Should `.github/workflows/yamllint.yml` be a path here to match what we're doing for JSON?

ripleymj: I think that's actually a typo from some of the testing I was doing out-of-band. We're not actually testing anything about the workflow.

PR #353, opened 2019-12-09
Move scripts and logs from /opt

kylelaker:
Since JMU Tux now lives in `/usr/share/icons` we can use that icon name and not have to keep it in `/opt`. The script also should live in `/usr/share`. And while we're moving stuff out, the Ansible log file can live in `/var/log`.

The scripts are also now templated by jinja2 to reduce the effort to change values of various paths later.


ripleymj:
This is a lot at once, and I may just be browsing too fast, but can you take a look at the Eclipse and jGRASP tasks as well. I believe they both expect `global_base_path` to download and unpack.

kylelaker:
Yeah it's a lot of reorganization. `global_base_path` still exists. It's now specified in `common` instead of `user` which probably makes a little more sense.

roles/robot-pkgs/tasks/main.yml
```diff
@@ -25,7 +25,7 @@
     path: '{{ global_profile_path }}'
     marker: "## {mark} ROS profile entries ##"
     block: |
-      source /opt/ros/{{ ros_release }}/setup.bash
+      source {{ global_base_path }}/ros/{{ ros_release }}/setup.bash
```
ripleymj: I'm not sure /opt should be templated here, since it comes from the apt package, not from Ansible.

roles/robot-pkgs/tasks/main.yml
```diff
@@ -46,7 +46,7 @@
     group: '{{ item.gid }}'
   loop: "{{ real_users }}"
 - name: Init Catkin workspace
-  shell: source /opt/ros/{{ ros_release }}/setup.bash && catkin_make
+  shell: source {{ global_base_path }}/ros/{{ ros_release }}/setup.bash && catkin_make
```
ripleymj: Same, about /opt

kylelaker:
Cool. Fixed the robots stuff

ripleymj:
Packer build fails with:
```
    virtualbox-iso: TASK [common : Install custom MintReport script] *******************************
    virtualbox-iso: fatal: [localhost]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'global_profile_path' is undefined"}
```
that's all I've got for now.

kylelaker:
Alright just did a run with the latest update. Packer ran and then I did a run as a user with all roles selected and it worked for me. Let me know if that seems to not be the case.

PR #360, opened 2019-12-17
Update to Tricia

### LINKED ISSUE #347: Update to Mint Tricia 19.3
New base release expected within weeks. Beta test branch available here: https://github.com/ripleymj/cs-vm-build/tree/tricia-beta-test

### LINKED ISSUE #359: packer build broken with VirtualBox 6.1: VBoxManage: error: Unknown option: --clipboard
It looks like the option has now become
`[--clipboard-mode disabled|hosttoguest|guesttohost|bidirectional]`.

Looks like we'll have to decide whether to build with 6.0 or 6.1 for Spring 2020.

ripleymj:
Assorted patches accumulated while beta testing Tricia. Holding this as a draft until mirror sync finishes, and I can run at least one build.

Closes #347 
Closes #359

### DIFF
--- a/.github/workflows/packerlint.yml
+++ b/.github/workflows/packerlint.yml
@@ -28,3 +28,4 @@ jobs:
       - name: Packer validate
         run: |
           $HOME/packer/pkg/packer_linux_amd64 validate packer/oem-build.json
+          $HOME/packer/pkg/packer_linux_amd64 validate packer/oem-build-beta.json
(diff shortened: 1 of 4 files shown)


kylelaker:
If we can use the `-var-file` argument to `packer` to do a beta build rather than requiring two nearly-identical files, that'd be awesome. If that requires too much refactoring work, then this solution works great.

Issue #339, opened 2019-09-28
jGRASP warns about unsupported Java distribution


kylelaker:
Somewhat similar to #154, we get a warning when starting jGRASP about unsupported Java; however, rather than being about the particular version, the warning is about the "vendor." In order to avoid the warning, we would need to either:

 - Install a "supported" install (Oracle, AdoptOpenJDK, etc.)
 - Use the jGRASP that bundles a JDK

Do we want to put in the effort to avoid this warning? If so I think just used the download with a bundled JDK would be easiest.

ripleymj:
jGRASP has confirmed privately that they will consider this OpenJDK build for inclusion in their next release, and will be adding a "Don't show warning again".

ripleymj:
jGRASP 2.0.6 beta 3 no longer warns about the Ubuntu JDK. Closing this in favor of #348 when it's released.

PR #364, opened 2020-01-10
Update jGRASP to 2.0.6

### LINKED ISSUE #348: Update to jGrasp 2.0.6
New release expected soon. Opening bug to track progress to spring image.

ripleymj:
Closes #348 

Need to run this at least once myself before merging.

### DIFF
--- a/roles/jgrasp/vars/main.yml
+++ b/roles/jgrasp/vars/main.yml
@@ -1,7 +1,7 @@
 ---
 # vars file for jgrasp
 jgrasp:
-  url: 'https://www.jgrasp.org/dl4g/jgrasp/jgrasp205_08.zip'
-  hash: '616c35b66d11dfb6e11e245e2e570c780efdd441'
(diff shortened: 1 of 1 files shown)


kylelaker:
#339 comes back with this PR.

kylelaker:
I guess I should be specific about the issue with #339. There is a "Don't Show Again" button but there is another, earlier warning, for which that is not an option. That warning will always appear every time that jGRASP is launched. It would make for an unfortunate user experience for CS1 students to see that message every time they launch the IDE for their class.

Should we consider using the version of jGRASP with an integrated JDK?

kylelaker:
Just want to note that the unsupported JDK warning persists even with this new re-released 2.0.6.

ripleymj:
I really do not want to balloon out a 7MB download to a 270MB download, just for an identical JVM. jGRASP had promised they would include the Ubuntu build but seem to have dropped that, hopefully accidentally. The labs exhibit the same behavior, so I'm not concerned about students being suddenly presented with this error.

PR #367, opened 2020-02-20
Update jGRASP to 2.0.6_02

kylelaker:
This will need to be backported as well. PR incoming.

### DIFF
--- a/roles/jgrasp/vars/main.yml
+++ b/roles/jgrasp/vars/main.yml
@@ -1,7 +1,7 @@
 ---
 # vars file for jgrasp
 jgrasp:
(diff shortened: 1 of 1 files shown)


ripleymj:
Does the VM GUI tool let you run branches with "/", or am I just typoing something?

kylelaker:
Looks it does not due to assumptions made in `branch_exists()`: https://github.com/jmunixusers/cs-vm-build/blob/master/roles/common/templates/uug_ansible_wrapper.py#L730.

I'll open a bug to track that.

ripleymj:
This appears to accept Ubuntu OpenJDK as legitimate now.

PR #374, opened 2020-03-31
Explicity label arguments to remove deprecation warning in Ubuntu 20.04

ripleymj:
Will add to this as I work through the rest of the warnings.

### DIFF
--- a/roles/common/templates/uug_ansible_wrapper.py
+++ b/roles/common/templates/uug_ansible_wrapper.py
@@ -139,9 +139,8 @@ class AnsibleWrapperWindow(Gtk.Window):
         contents.set_border_width(10)
 
         label = Gtk.Label(label=
-            "Select the course configurations to add/update"
-            " (at this time courses cannot be removed)."
-        )
+                          "Select the course configurations to add/update"
+                          " (at this time courses cannot be removed).")
         label.set_alignment(0.0, 0.0)
 
         contents.pack_start(label, False, False, 0)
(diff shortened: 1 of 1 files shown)


roles/common/templates/uug_ansible_wrapper.py
```diff
@@ -138,18 +138,17 @@ def __init__(self):
         contents = Gtk.Box(orientation=Gtk.Orientation.VERTICAL, spacing=12)
         contents.set_border_width(10)
 
-        label = Gtk.Label(
-            "Select the course configurations to add/update"
-            " (at this time courses cannot be removed)."
-        )
+        label = Gtk.Label(label=
+                          "Select the course configurations to add/update"
+                          " (at this time courses cannot be removed).")
```
kylelaker: This is unlikely to be syntactically valid. You'll need to wrap the strings in `()` so that Python actually knows they're part of the same expression. Alternatively, we'll need to just accept really long lines in some places. Or move to a better method of string management.

```suggestion
        label = Gtk.Label(
            label=(
                "Select the course configurations to add/update"
                " (at this time courses cannot be removed)."
            )
        )
```

kylelaker: For clarification, this only applies to strings that span multiple lines.

kylelaker:
I added a few more fixes for the `DeprecationWarning`s but that `spawn_async` is going to be interesting. I don't even really want to start thinking about that one.

kylelaker:
Do we want to merge this as-is? Or are we going to mandate a `spawn_async` fix?

ripleymj:
I'm ok merging as is, especially since we don't have it set to close the parent issue. I can't review since I own the PR though, so if you, @Advill, or @nuttercd want to do the honors, please do.

kylelaker:
Looks like we've got enough approvals. I'll go ahead and merge. Thanks, all!

PR #378, opened 2020-05-01
Remove CS 354 (Robotics) support

### LINKED ISSUE #377: RFC: Deprecate CS354 support
Our CS354 robotics role has always been in an awkward place, since it was never formally requested, has no means of verification, has no known users, and is for a course only taught occasionally. For the fall image, I would like to remove the CS354 option from the setup GUI and stop testing it on a routine basis. I am not proposing to delete the role from the repository at this point, though I'd discourage anyone from spending significant time maintaining it. We can certainly bring it back later, though I think it would need to have a set of requirements defined, and a sponsor to test and validate correctness of the role.

kylelaker:
This role has long been under-tested compared to the rest of the
offerings and it currently does not have any known users. The course is
offered rather infrequently and the students in the course and club
generally seem to prefer the lab machines that are provided.

This removal is intended to ensure that there is not code rotting on the
master branch. It is not intended to backport this removal to any prior
release.

This PR is being opened to provide a possible implementation of the
RFC in issue #377. It will remain in draft status until a direction is agreed
upon.

Closes #377


ripleymj:
Go ahead and move this to Ready. I was only going to push for dropping 354 in the GUI, leaving the option to ansible-pull it directly, but nobody has spoken up for it, so let's kill it all.

kylelaker:
I guess that is an important part of the removal. And the README doesn't really show up in `grep -i -r "CS354" roles/` :man_shrugging: Hopefully we're all set now

PR #379, opened 2020-05-01
Update jGRASP to 2.0.6_03

kylelaker:
Let me know if we want to backport to the spring image.


ripleymj:
I guess we should since we haven't published a newer image

PR #388, opened 2020-05-22
Add linter to check hashes

kylelaker:
I have no idea how to properly test this. The goal is to have a linter that will be able to validate the hashes and URLs for the `vars` files. 

In my mind, this should just be a quicker way to validate that the new hashes and URLs that are given in a PR are valid without having to spin the VM up to validate them. We may accidentally get build failures if software updates and someone is modifying one of the listed files for an unrelated reason. I'd argue that if you're modifying one of the files that this PR watches, fixing hashes and URLs as necessary would generally be in-scope for that PR.

### DIFF
--- a/scripts/hashlint.py
+++ b/scripts/hashlint.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python3
+
+"""
+Simple script to validate the various hashes for downloads across the project.
+"""
+
+import hashlib
+import sys
+
+import requests
+import yaml
+
+# https://requests.readthedocs.io/en/master/user/quickstart/#timeouts
+REQUEST_TIMEOUT_SEC = 5
+URLS = {
+    "roles/jgrasp/vars/main.yml": {
+        "hash": "jgrasp.hash",
+        "urls": ["jgrasp.url"],
+    },
+    "roles/finch/vars/main.yml": {
+        "hash": "finch.hash",
+        "urls": ["finch.url"],
+    },
+    "roles/eclipse/vars/main.yml": {
+        "hash": "eclipse.hash",
+        "urls": ["eclipse.url", "eclipse.url_backup"],
+    },
+}
+
+
+def get_field(data, key):
+    """
+    Get a field separated by dots
+    """
+
+    if isinstance(key, str):
+        key = key.split('.')
+    else:
+        print(type(key))
+    while key:
+        data = data[key.pop(0)]
+    return data
+
+
+def check_software_hash(session, url, expected, file):
+    """
+    Checks the hash for the contents of a URL against an expected value
+    """
+
+    try:
+        response = session.get(url, timeout=REQUEST_TIMEOUT_SEC)
+        response.raise_for_status()
+        data = response.content
+    # pylint: disable=broad-except
+    except Exception:
+        print(f"{file}: Unable to download {url}", file=sys.stderr)
+        return False
+
+    sha1 = hashlib.sha1(data).hexdigest()
+    if sha1 != expected:
+        print(f"{file}: Expected {expected}. Found {sha1}", file=sys.stderr)
+        return False
+
+    print(f"{file}: Validated {sha1=} as expected")
+
+    return True
+
+
+def main():
+    """
+    Main
+    """
+
+    session = requests.Session()
+    # User-Agent is the same as used by Ansible itself
+    # https://github.com/ansible/ansible/blob/062e780a68f9acd2ee6f824f252458b8a0351f24/lib/ansible/modules/get_url.py#L167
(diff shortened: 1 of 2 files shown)


.github/workflows/hashlint.yml
```diff
@@ -0,0 +1,29 @@
+---
+name: Python lint workflow
```
ripleymj, line 2: Can you make this something like "Hash lint workflow" so it shows differently on the actions dashboard?

.github/workflows/hashlint.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Set up Python 3.x
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.x
+      - name: Install Python dependencies
+        run: |
+          pip3 install -U pyyaml requests
+      # We cannot just do roles/*/*/*py because we do not want to lint the
```
ripleymj, line 25: This seems like it got copied from another file :)

scripts/hashlint.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python3
```
ripleymj, line 1: Could you add this script to the Python lint workflow?

ripleymj:
In a broader sense, I'm not sure if we should have this attached to the pull_request action, or if it should be a weekly schedule, and open an issue if it fails. I'm ok with doing both, as long as failure on a PR doesn't spawn a new issue.

.github/workflows/hashlint.yml
```diff
@@ -0,0 +1,29 @@
...
+      # finch scripts which we cannot change
+      - name: Run common role pylint
```
ripleymj, line 27: Sorry, naming here too please

.github/workflows/hashlint.yml
kylelaker: You caught me :)

scripts/hashlint.py
kylelaker: I thought it'd be caught by `**.py` but I can add this one explicitly if needed

ripleymj: Just checked that workflow file... we trigger on **py but only lint `roles/*/*py`... so that's another thing :(

kylelaker:
>In a broader sense, I'm not sure if we should have this attached to the pull_request action, or if it should be a weekly schedule, and open an issue if it fails. I'm ok with doing both, as long as failure on a PR doesn't spawn a new issue.

That's far beyond my current GitHub Actions knowledge (don't let that surprise you since I had to copy `pylint.yaml` to get this PR opened :). I can look into that but this seemed like a small victory we could claim at the moment. If you have suggestions of where to look, let me know.

I am also happy to add subscribers to the SNS topic written to by my [script](https://github.com/kylelaker/updatechecker) if that'd be helpful while this feature is worked out.

scripts/hashlint.py
```diff
@@ -0,0 +1,88 @@
...
+
+    return True
+
+
+def main():
+    """
+    Main
+    """
+
+    session = requests.Session()
```
ripleymj, line 69: I was going to ask if there's a reason to recycle session between requests, but since we're not using auth or cookies, is a session even necessary?

scripts/hashlint.py
```diff
@@ -0,0 +1,88 @@
...
+    return data
+
+
+def check_software_hash(session, url, expected, file):
+    """
+    Checks the hash for the contents of a URL against an expected value
+    """
+
+    try:
+        data = session.get(url).content
```
ripleymj, line 50: Maybe borrow Ansible's user-agent to appease Finch: https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/get_url.py#L170

scripts/hashlint.py
kylelaker: If we apply the suggestion for the 'User-Agent' header, we can do it on the session instead of each request. A session isn't really required though.

scripts/hashlint.py
kylelaker: I'll clean that up while I'm in there.

kylelaker:
Holding off on any changes until #392 is merged.

kylelaker:
Alright, so this is a rev 1. It doesn't do any sort of periodic running, it _just_ validates the hashes in a PR/master. I think converting to run periodically and getting it to open an issue on failure might be a great first issue for a new contributor (but I can add it myself later if we don't find one).

Seems like it can take awhile to run if we get a slow mirror for Eclipse. It certainly would be interesting to see at some point how long that step takes. Our end users probably see similar behavior.

PR #392, opened 2020-06-01
Unify lint tests into one workflow, one badge

### LINKED ISSUE #376: Upgrade Actions to setup-python@v2
New Python setup utility has been released by GitHub, and has better support for pre-cached environments. Let's evaluate and upgrade our testing.

ripleymj:
Trying to clean up our initial attempt at Actions.

Needs to be integrated with #388 
Closes #376


ripleymj:
Moving this back to a draft. I got a reminder this morning that Ubuntu 20.04 test images are available, even if the ubuntu-latest target hasn't been moved forward. I'd like to try a matrix build against both 18 and 20 before merging this.

ripleymj:
Scratch that, 20.04 is still in pre-release and I think Actions will upgrade -latest right about the time we're planning to move. Not worth the doubled runtime to pursue at this point, IMO.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,91 @@
+---
+name: VM lint workflow
+on:
+  pull_request:
+  push:
+    branches:
+      - master
```
kylelaker, line 7: Should the various release branches be included? `**a` might catch any valid branches we care about. Alternatively, we could just do all branches.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,91 @@
+---
+name: VM lint workflow
+on:
+  pull_request:
```
kylelaker, line 4: I always forget that YAML allows this syntax. Is this typical of GitHub Actions definitions? 

```suggestion
  pull_request: null
```

is a little more explicit. They should parse to the same thing.

ripleymj: Yeah, I did it as they demonstrate it: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#on and haven't seen yamllint complain

.github/workflows/lint.yml
ripleymj: I don't think so. I just added push-master for the README badge, so it wouldn't reflect the status of the last PR touched. I don't think our history branches are generally lint clean at this point, and may get worse when tested against a 20.04 base

PR #396, opened 2020-06-16
Ulyana fixes v1

### LINKED ISSUE #384: Add output_filename support
Add this once it's in a released Packer version, so image-$semester.ova files are made automatically.

https://github.com/hashicorp/packer/pull/9174

ripleymj:
Assorted updates to clean up in preparation for a new version.

Closes #384


packer/mint-build.json
```diff
@@ -112,7 +113,8 @@
   "post-processors": [
     {
       "type": "checksum",
-      "output": "{{user `output_dir`}}/{{user `vm_name`}} {{user `semester`}}.{{.ChecksumType}}sum"
+      "checksum_types": "sha256",
+      "output": "{{user `output_dir`}}/image-{{user `semester` | lower }}.{{.ChecksumType}}sum"
```
kylelaker, line 117: Can this be

```
"output": "{{user `output_dir`}}/{{user `output_filename`}}.{{.ChecksumType}}sum"
```

Seems like if it can, Ubuntu would need the same change. Don't know Packer well so just ignore if not doable.

ripleymj: I don't think so, looks like `output_filename` is a parameter not a variable

PR #399, opened 2020-06-27
Update Packer lint to use new packaging format

ripleymj:
Packer seems to have updating their nightly packaging, breaking our lint test. I can't find any documentation that confirms the change was intentional/permanent, but I'm drafting this in case the change stays.


ripleymj:
This packaging change has persisted through another nightly build, so I'm recommending this for merging now.

PR #402, opened 2020-07-13
Update snaps during Ubuntu OEM install

### LINKED ISSUE #385: Update snaps during OEM
Ubuntu 20.04 seems more reliant on default-installed snaps. We should update them during OEM so there's less to do at first boot. Usage in Mint is unknown at this point, but they should be updated too.

ripleymj:
Nothing exciting, just want to ... 

Close #385


kylelaker:
Of course the `snap` module doesn't have a `refresh` option...

PR #389, opened 2020-05-30
Remove spawn_async warning

kylelaker:
Since Focal is the main focus now, this should be fine. It won't work on 19.3 because `spawn_async` isn't available in that version Vte.

Resolves #373

PR #403, opened 2020-07-13
Use 20.04 specifically for Actions

kylelaker:
20.04 is _technically_ still in "preview" but even when it's not, we're
primarily focused on the 20.04 release. So we also want to guard against
when `latest` is actually 22.04 or whatever.


kylelaker:
This _should_ also make #389 validate correctly.

PR #405, opened 2020-07-13
Install python wheel package

kylelaker:
This should clean up some messages during the Python linting and allow
packages to be installed using the wheel format rather than the "legacy"
setup.py method.


kylelaker:
Doesn't fix it. And not really worth diving into further.

PR #407, opened 2020-08-10
Add cs432

lam2mo:
This adds (or attempts to add) CS432 support. I've ported the major class projects to C (previously they were in Java) and now the supported environment is nearly identical to CS261 (e.g., build/test with gcc/gdb/make/check, and need to work on stu or another Ubuntu-based machine because of compiled binaries and Valgrind).

The one difference that I can think of is that it's useful in CS432 to be able to build graphical ASTs using the "dot" utility. This is somewhat optional, but generally a useful thing and it's not too large so I've just added it (via the Graphviz package) to the "advanced programming" role for CS261/361/432.

I have not tested this myself so this is really just a suggested change and basis for discussion at this point. I'm not really set up to build it at the moment (getting an error about having too low of a Packer version) but I'm happy to test an image if someone else can build it.

Comments/questions welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,9 @@ using these tasks and the VM is available
 * Prolog
   * swi-prolog
 
+### CS432
+* (essentially same as CS261)
+
 ## Manual use of these Ansible roles
 
 This project was developed on and for Linux Mint, and receives occasional
--- a/README.md
+++ b/README.md
@@ -51,9 +51,9 @@ using these tasks and the VM is available
     * g++
     * gcc
     * gdb
+    * graphviz
     * logisim
     * valgrind
-    * graphviz
   * Code editors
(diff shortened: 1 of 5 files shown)


lam2mo:
Oh, and I fixed a typo. :)

lam2mo:
Oh, and the Packer lint check is failing so that should be addressed. Pretty sure that's unrelated to this pull request, though.

lam2mo:
Huh. Didn't even realize they were alphabetized.

ripleymj:
With #410 merged, if you rebase, I think lint checks will pass. I'd like to run it at least once before merging, but everything looks good so far.

lam2mo:
Jedi operation complete. :)

PR #418, opened 2020-09-21
Add action step to test all roles

kylelaker:
This will validate that the roles at least run. It does not
necessarily validate any sort of correctness.


nuttercd (dismissed):

kylelaker:
Going to close this. If we decide we want it, we can always consider reopening it. But I don't want to just keep a stale PR around.

PR #419, opened 2020-09-21
Switch to supporting VirtualBox EFI mode

kylelaker:


kylelaker:
My remaining concerns with this are:
 - Whether VirtualBox EFI support is mature enough
 - Setting the `d-i grub-installer` values in `oem-preseed.cfg`
 - Setting the debconf `grub-pc` values for `install_devices` and `install_devices_disks_changed`

PR #422, opened 2020-10-30
Updated role names & implemented Ansible lint rules

### LINKED ISSUE #415: Properly handle new Ansible lint rules
As seen in #413's tests, there are some new Ansible Lint rules that we are violating. Rather than just warning as we do with #414, a proper resolution (either skipping or resolving) the rules should be implemented.

To completely resolve this, the following will need to be completed (PRs that implement only a subset of these _will_ of course be accepted):
- [ ] Rename all roles using kebab case (ex: `adv-prog-pkgs`) to use snake case (ex: `adv_prog_pkgs`), basically, replace all hyphens (`-`) in role names with underscores (`_`)
- [ ] Update `local.yml` and `oem.yml` with the new role names
- [ ] Undo the changes made in #414
- [ ] Ensure that a `mode` is specified any time a file or directory is written or created

Bronstrom:
Closes #415


roles/user/tasks/main.yml
```diff
@@ -4,10 +4,16 @@
...
     state: directory
+    owner: root
+    group: root
+    mode: 0755
 - name: Create local applications directory
   file:
     path: /usr/local/share/applications/
     state: directory
+    owner: root
+    group: root
```
ripleymj, line 15: This seems to have been group "staff" before, but everything else in that hierarchy is owned by "root". Just noting the change, but I think it's better this way.

roles/finch/tasks/main.yml
```diff
@@ -19,19 +22,31 @@
...
+    owner: root
+    group: root
+    mode: 0644
 - name: Install Finch extra - dance2.py
   copy:
     src: dance2.py
     dest: '{{ finch.install_path }}'
+    owner: root
+    group: root
+    mode: 0755
```
ripleymj, line 42: Without the shebang, these aren't actually executable, so I think I was mistaken in telling you 0755 and it should probably be "0644". Thoughts @kylelaker?

roles/finch/tasks/main.yml
```diff
@@ -19,19 +22,31 @@
...
 - name: Install Finch extra - maze.py
   copy:
     src: maze.py
     dest: '{{ finch.install_path }}'
+    owner: root
+    group: root
+    mode: 0755
```
ripleymj, line 49: Same as dance2.py

roles/finch/tasks/main.yml
kylelaker: Any Python module that won't be executed (via a call to `execve(2)` or friends) does not need `0755` and should be `0644`. If a Pythin file has a hashbang or `if __name__ == '__main__'`, it's probably intended to be `execve(2)`-ed. Without a hashbang and with `0755` if you were to `./dance2.py`, bash would run it as bash which isn't ideal. We should make sure of this across all the Finch scripts.

roles/user/tasks/main.yml
kylelaker: With `0755` there also isn't a difference I'm aware of for a typical user between its group being `staff` vs `root`, even when a member of `staff`. On my Debian box, the only thing that isn't `0755` and `root:root` in `/usr/local/share` is `fonts/` which is `2775` and `root:staff`.

On a fresh Linux Mint, `/usr/local/share/applications/` doesn't seem to exist, so `0755` and `root:root` feels like the best option until we're proven otherwise.

ripleymj (approved):

kylelaker (approved):

PR #428, opened 2020-12-16
Revert "Stop trying to install transitional chromium package"

### LINKED ISSUE #394: Address Chromium "situation"
Chromium as packaged is intentionally unusable in Mint 20. I'm working on another PR that removes it from the current playbook, so we don't give the wrong impression of it working. The alternatives are somewhat documented here: https://forums.linuxmint.com/viewtopic.php?f=42&t=321953 or we can close this with an intentional WONTFIX on record.

ripleymj:
Now that Ubuntu has a `chromium-browser` package that installs a snap, and Mint has a native package to provide the browser, we should be able to use this as it once was.

This reverts commit 8e0b9b0c88884d55d3e97babbd6cbb97e735f766.

Closes #394

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ using these tasks and the VM is available
 * Mint packages (basic-prog-pkgs role)
   * artha
   * bless
+  * chromium-browser
   * geany
   * idle
   * libreoffice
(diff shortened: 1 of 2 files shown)


kylelaker (approved):
Only concern I had was whether this would break idempotency. I thought we removed `snapd` at some point but I misremembered; we just [refresh snaps](https://github.com/jmunixusers/cs-vm-build/blob/main/roles/oem/tasks/ubuntu_only.yml) on Ubuntu only.  Glad we can get this back in now.

PR #431, opened 2021-01-03
Add script to automate packer testing

kylelaker:
This script will, given a PR ID and a template file, automatically
generate a packer var file and invoke packer to test a PR. It is
important to note that it is still a requirement that the code actually
be pushed to the branch (testing changes that are only local work just
as ineffectively as before) but at least it's (possibly) moderately
easier to test than with manually fetching the git info.

As an additional helper, since testing may happen on more platforms than
traditional building of the VM does, an attempt is made to guess the
proper audio device type for the host OS. If it doesn't work, then we
just give up and either default to "null" or prompt the user (if we're
running in interactive mode).

Some example invocations:

```
../scripts/packer_pr_test.py --var-file beta-vars.json 430 mint-build.json
```

or

```
../scripts/packer_pr_test.py 430 ubuntu-build.json
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -61,7 +61,7 @@ jobs:
           sudo apt-get -q install python3-gi-cairo gir1.2-gtk-3.0 libgirepository1.0-dev gir1.2-vte-2.91
       - name: Install dependencies
         run: |
-          pip3 install pyyaml pygobject pylint requests
+          pip3 install pyyaml pygobject pylint requests click PyGithub
       # We cannot just do roles/*/*/*py because we do not want to lint the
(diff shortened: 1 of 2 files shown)


kylelaker:
Linting failure is a [Pylint bug](https://github.com/PyCQA/pylint/issues/3882) on 3.9. [`typing.Optional`](https://docs.python.org/3/library/typing.html#typing.Optional) _is_ subscriptable. Should we be listing on the latest version of Python anyway? Or something closer to the latest Mint/Ubuntu release? In the interim, this PR can either sit (or be rejected) or the type hints can be removed.

ripleymj (approved):

PR #432, opened 2021-01-03
Restore stu-home files

kylelaker:
In contrast to departmental guidance in 2018, it seems that it is now [supported](https://wiki.cs.jmu.edu/student/utilities/sshfs) (and recommended) to access stu via sshfs support.

Technically this isn't sshfs; it's Nemo's sftp support. But it's all basically the same and will largely present the same to students.

This partially reverts commit 190e9136e8cfc37eea2c5e3e88a303a12cb4d2f7.
This reverts commit 8bb25748cc87b5707fc15254f3984ddc6463877f.

### DIFF
--- a/roles/user/files/stu-home.desktop
+++ b/roles/user/files/stu-home.desktop
@@ -1,0 +1,10 @@
+[Desktop Entry]
+Version=1.0
+Name=Home on Stu
+Exec=sh -c "nemo sftp://\${USER}@stu.cs.jmu.edu/cs/home/stu/\${USER}/"
+Comment=Mount/umount your JMUCS NFS home directory
+Terminal=false
+Type=Application
+Categories=System;
+Icon=usermount
+Name[en_US]=Home on Stu
--- a/roles/user/files/stu-home.desktop
+++ b/roles/user/files/stu-home.desktop
@@ -6,5 +6,5 @@ Comment=Mount/umount your JMUCS NFS home directory
 Terminal=false
 Type=Application
 Categories=System;
-Icon=usermount
+Icon=gnome-fs-ssh
 Name[en_US]=Home on Stu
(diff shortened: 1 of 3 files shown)


ripleymj:
This is still probably as bad an idea as it ever was, but I do suppose it puts us in line with other platforms. When I run it, I get a nondescript text document icon on the desktop. Do you think we be possible to get something more network or disk related? Also curious how you'd feel about having the Ansible known_hosts module pre-push the stu key to prevent a nag on first connection?

kylelaker:
Alright @ripleymj I think that's done. The icon is updated to  image and I have tried to have it implicitly trust `stu.cs` but I can't test the whole login flow since my account is dead.

ripleymj:
Purely for style, adding `hash_host: yes` will make the entry look like others.

As an actual problem, the known_hosts file is being created with root ownership and preventing any future keys from being added.

kylelaker:
Eww. How do we want to fix that? Would a `{ file: {state: present}}` with the right owner before/after be good enough? Assuming they both work, do you have a preference for before or after?

ripleymj:
If we go that route, I think after would be safer to make sure we can't leave it in a bad state. Otherwise become/become_user might be the other option

kylelaker:
Rather than touching every user's config, why don't we just rely on setting it globally? Apparently there's already a decent default for `GlobalKnownHostsFile` per `ssh_config(5)`. At some point we'll want to add some sort of linting/validation that the host key for `stu.cs` hasn't changed but that also probably happens on a predictable cadence.

ripleymj:
That works, and there will be screaming everywhere if that key changes. Do you want to remove creation of the $HOME/.ssh then, since we're not using it?

kylelaker:
Done 👍

ripleymj (approved):

PR #433, opened 2021-01-08
Update Mint build to Ulyssa release

ripleymj:
Going to leave this as a draft until a test build completes, but no issues were found with the beta.

PR #434, opened 2021-01-15
Ensure jGRASP wedge gets built and installed

kylelaker:
Resolves an issue uncovered during #433 testing

### DIFF
--- a/roles/jgrasp/tasks/main.yml
+++ b/roles/jgrasp/tasks/main.yml
@@ -2,12 +2,16 @@
 # tasks file for jgrasp
 - name: Install jGRASP dependencies
   apt:
-    name: lsb-core
+    name: '{{ jgrasp_wedge_dependencies }}'
     state: latest
-- name: Check jGRASP
+- name: Check jGRASP zip
   stat:
     path: '{{ jgrasp.zip }}'
-  register: st
+  register: zip
+- name: Check jGRASP wedge
+  stat:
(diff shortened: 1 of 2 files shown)


kylelaker:
This method prevents having `/opt/jgrasp/internal_bin` needing to remain world-writable. I hate a `./configure && make` but that's what jGRASP would do anyway behind the scenes. From the docs:

> ## Installing on Linux
> Unzip the distribution file in the directory where you wish to install jGRASP. This will create a jgrasp directory containing all the files. You may need to set executable permissions on all files, using chmod -R 777 jgrasp (777 is read/write/execute permissions for all - change as appropriate). You may want to add the "bin" subdirectory to your execution path or create a soft link to .../jgrasp/bin/jgrasp from a directory on the executable path.

> The bin/create_linux_launcher.sh script will create a desktop launcher for jGRASP on most Linux systems.

> Note that java must be on your system path in order to run jGRASP.

> The first time you compile or run a program jGRASP will attempt to build the native "wedge" and startup shell programs. This may fail due to file permission issues or for other reasons, in which case jGRASP will use a less capable pure Java wedge. Without the native wedge, when you run a Java program I/O will usually be line buffered, and when you run a C or C++ program I/O will usually be block buffered (which is useless for interactive command line I/O). To build the native parts from the command line, cd to the jGRASP "src" directory. Run ./configure and if successful run ./Make.sh .

> Please contact us if building the native "wedge" fails for some reason other than a C compiler not being installed.

ripleymj (approved):

Issue #436, opened 2021-02-05
Update Ubuntu image


ripleymj:
Ubuntu 20.04.2 has been released, so we probably need to update to keep Packer Actions happy, like #410 

https://lists.ubuntu.com/archives/ubuntu-announce/2021-February/000264.html

Issue #409, opened 2020-08-11
Prevent Packer breakage when a new ISO is released (or fix current breakage)


kylelaker:
As seen in the checks for #407, when a new Ubuntu ISO is released and the `ubuntu-build.json` file still references the old ISO. This no longer exists [on the mirror](https://mirror.cs.jmu.edu/pub/ubuntu-iso/focal/). Therefore, it's not referenced in `SHA256SUMS` which results in a linter failure.

Since we run all linters on each PR, this can result in failures that aren't expected.

Ideally, we'd work up a good long-term fix for this. If that's not really feasible then just fixing it this time and remembering that check can fail sometimes is a sufficient alternative.

The new ISO file is `ubuntu-20.04.1-desktop-amd64.iso`.

PR #437, opened 2021-02-06
Automatically update Ubuntu ISO

kylelaker:
This adds the ability to automatically update the Ubuntu ISO within a release train which is
specified in the configuration. This will check every day at 9AM or when invoked manually. It will
push to an `ubuntu-update/` branch and open a pull request if there is a new ISO available.

Resolves: #436
Resolves: #409

### DIFF
--- a/.github/workflows/auto-update-iso.yml
+++ b/.github/workflows/auto-update-iso.yml
@@ -1,0 +1,42 @@
+---
+name: Update ISOs
+on:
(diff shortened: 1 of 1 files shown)


kylelaker:
For an example, see https://github.com/kylelaker/test-actions-temp which I'll keep around at least until this PR is merged

ripleymj (approved):

PR #440, opened 2021-02-12
Update for Ansible Lint 5

ripleymj:
Ansible Lint 5 doesn't seem to automatically depend on Ansible via pip, so adding that explicitly now. Also switches to named rules instead of numeric, causing an additional linting failure for us.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,4 +1,3 @@
 skip_list:
 # Package installs should not use latest
-  - 403
-
+  - package-latest
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ jobs:
           python-version: 3.x
       - name: Install dependencies
         run: |
-          pip3 install ansible-lint
+          pip3 install ansible ansible-lint
       - name: Run ansible-lint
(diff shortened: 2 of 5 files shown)


kylelaker (changes requested):
Looks like `.ansible-lint` is where we disable things globally, but we selectively skip a few checks inline.

```
$ ansible-lint
Added ANSIBLE_ROLES_PATH=roles
[WARNING]: log file at /var/log/vmtools/ansible_runs.log is not writeable and we cannot create it, aborting

WARNING  Replaced deprecated tag '306' with 'risky-shell-pipe' but it will become an error in the future.
WARNING  Replaced deprecated tag '301' with 'no-changed-when' but it will become an error in the future.
```

```
$ grep -r "noqa" roles 
roles/oem/tasks/mint_only.yml:- name: Upgrade all installed packages  # noqa 301
roles/oem/tasks/ubuntu_only.yml:- name: Upgrade all installed snaps  # noqa 301
roles/oem/tasks/vm_only_post.yml:- name: Validate missing USB 2/3 controllers  # noqa 306
```

kylelaker (approved):

PR #442, opened 2021-02-16
Cleanup packer var specifications

kylelaker:
In hashicorp/packer#7390, the ability to reference a user variable
within a user variable was added. This greatly simplifies the definition
of the version and reduces some duplication there.

### DIFF
--- a/packer/beta-vars.json
+++ b/packer/beta-vars.json
@@ -1,10 +1,10 @@
 {
   "semester": "Sp21",
-  "mint_version": "20.1",
+  "version": "20.1",
 
   "git_repo": "https://github.com/jmunixusers/cs-vm-build",
   "git_branch": "main",
 
   "mirror_url": "http://mirror.cs.jmu.edu/pub/linuxmint/images/testing",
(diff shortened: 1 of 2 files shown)


packer/edge-vars.json
```diff
@@ -0,0 +1,3 @@
+{
+    "iso_file": "linuxmint-{{user `version`}}-cinnamon-64bit-edge.iso"
```
ripleymj, line 2: Do we really want to introduce the edge builds into this? I don't even think we can run successfully on them at this point, and VirtualBox is unlikely to give us the exotic hardware its intended for.

ripleymj (changes requested):

packer/edge-vars.json
kylelaker: A quick run worked for me using edge but yeah it probably doesn't provide much value. Included it as an example of being able to reference `version` in another file but it's not particularly useful. I'll push another change later today with it removed.

ripleymj (approved):

PR #443, opened 2021-02-16
Add custom Ansible lint rule for mode arguments

kylelaker:
Using non-string modes can be unpredictable in atypical cases. It is
usually safer to just use strings, whether representing the typical
numeric representation of permissions or the symbolic representation.

This is a lint check to enforce the opinions from #441.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,7 +20,7 @@ jobs:
           pip3 install ansible ansible-lint
       - name: Run ansible-lint
         run: |
-          ansible-lint local.yml oem.yml
+          ansible-lint -R -r lintrules/ansible local.yml oem.yml
   Packer:
     name: Run Packer lint tests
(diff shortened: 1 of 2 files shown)


ripleymj:
Do you have any examples of why quoted vs leading zero is better? Both the Ansible and Ansible Lint docs suggest they're equivalent, and we only noted it as a bug in 2.1.

kylelaker:
The docs say:

> You must either add a leading zero so that Ansible's YAML parser knows it is an octal number (like 0644 or 01777) or quote it (like '644' or '1777') so Ansible receives a string and can do its own conversion from string into number.

So just to enforce consistency and prevent incorrect interpretation. `directory_mode` still gives warnings about how its parsing it as `493` or whatever. Seems like the kind of detail that can get missed in review and checking for a leading `0` is harder in a lint rule since the YAML has already been parsed :)

ripleymj (approved):

PR #447, opened 2021-02-21
Remove orange badge from mintupdate icon

kylelaker:
On first boot (and until the user clicks "OK" on the welcome page), the
mintupdate-launcher icon has an orange badge on it. Since this tool
handles installing updates, it's showing an orange badge just to have
users press "OK" and see no updates.


kylelaker:
(Another option would instead by to set `hide-systray` to `true` but for that we might want to _only_ do it in OEM)

ripleymj (approved):

PR #449, opened 2021-02-26
Update FinchPython120 hash

kylelaker:
Yet again Bird Brain Technologies has pushed out an updated zip file
with a new SHA1 hash without bumping any sort of version number. That
breaks the downloads. This updates the hash to fix it.


kylelaker:
Oof @ripleymj how do you want to handle these failures? We don't typically backport the pipeline stuff...

ripleymj:
Pretty sure Actions on all our previous branches are broken too. It's a targeted fix that just tested against `main`, so I'm not really worried.

ripleymj (approved):

PR #453, opened 2021-04-12
Basic ARM support

kylelaker:
This gets the configuration scripts working well enough on Ubuntu
aarch64 that there aren't failures for any role and that it doesn't
break any core parts of the OS.

It's not perfect since the y86 tools and others aren't ready for
aarch64. I don't really intend to propose that we go all-out on
making aarch64 work perfectly; just that we don't break too much that
already works if someone does try to run it.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -61,7 +61,7 @@ jobs:
           sudo apt-get -q install python3-gi-cairo gir1.2-gtk-3.0 libgirepository1.0-dev gir1.2-vte-2.91
       - name: Install dependencies
         run: |
-          pip3 install pyyaml pygobject pylint requests click PyGithub
+          pip3 install pyyaml pygobject pylint click PyGithub jinja2 aiohttp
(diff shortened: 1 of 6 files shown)


kylelaker:
lol at the hash validation failure. can hack out a solution to that if this PR is even something we want

ripleymj:
I think we're in agreement about the uncertainty of ARM support, but I'm ok with small targeted fixes like this that allow us to run on ARM if someone is adventurous and don't complicate the AMD64 path. If you have ideas on supporting the dual-arch hash check, I guess I'd prefer to see that merged alongside this.

kylelaker:
Looks like `mirror.cc.columbia.edu` is down at the moment (getting timeouts for ping, tcp/80, and tcp/443)  so we'll need to re-run the checks at a later time; but in a way it's convenient because we can at least see what one type of failure looks like and validate it still fails on failure :)

ripleymj (approved):

PR #454, opened 2021-05-31
Misc fixes may21

ripleymj:
Just a few cleanups discovered in playing with the codebase lately.

1. Eclipse Checkstyle has moved the repository, so update that URL
2. The stu-home shortcut uses Cinnamon Nemo and doesn't work with Ubuntu, so make that a mint-only task. Ubuntu shortcuts also seem to require world-exec and `gio set file.desktop "metadata::trusted" yes` to run
3. Ubuntu 21.04 packer aborts without the apt "-y" flag, which we use elsewhere and most examples seem to show. I'm not sure how this was working before, but I think noninteractive mode is meant to suppress package prompts
4. The Ubuntu ISO checker is running on forks, which would likely cause diverging commits on the main branch

### DIFF
--- a/.github/workflows/auto-update-iso.yml
+++ b/.github/workflows/auto-update-iso.yml
@@ -8,6 +8,7 @@ env:
   UBUNTU_RELEASE: focal
 jobs:
   update-ubuntu:
+    if: github.repository == 'jmunixusers/cs-vm-build'
     name: Update Ubuntu ISO file name
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


kylelaker (approved):
Thanks for catching these. Definitely have no idea how `apt` was ever working before. We definitely shouldn't have been expecting it to. Good stuff.

kylelaker:
Looks like `ansible-lint` is now grumpy about some unrelated things (but just grumpy; not failing) with `ignore_errors`. We can track this separately.

PR #459, opened 2021-06-12
Space savings

ripleymj:
Add a few more package removals to save about 100MB and add snap cleanup trick from StackOverflow to save about 400MB in Ubuntu builds.

### DIFF
--- a/roles/oem/tasks/ubuntu_only.yml
+++ b/roles/oem/tasks/ubuntu_only.yml
@@ -6,6 +6,22 @@
     upgrade: safe
 - name: Upgrade all installed snaps  # noqa no-changed-when
   command: snap refresh
+- name: Purge old snaps
+  shell: |
+    set -o pipefail
+    errors=0
+    changed=""
+    while read snapname revision; do
+      snap remove "$snapname" --revision="$revision"
+      errors=$((errors + $?))
+      changed="Snap removed"
+    done <<< "$(snap list --all | awk '/disabled/{print $1, $3}')"
+    echo $changed
+    exit $errors
+  args:
+    executable: /bin/bash
+  register: snapout
+  changed_when: '"Snap removed" in snapout.stdout'
 - name: Copy dconf config file
   copy:
     src: csconfig-ubuntuonly
--- a/roles/oem/vars/main.yml
+++ b/roles/oem/vars/main.yml
@@ -41,6 +41,7 @@ packages_to_remove:
   - fonts-tibetan-machine
   - fonts-tlwg-*
   - fonts-wqy-microhei
+  - fonts-yrsa-rasa
   - hexchat*
   - hunspell-de-*
   - hunspell-en-au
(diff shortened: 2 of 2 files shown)


roles/oem/tasks/ubuntu_only.yml
```diff
@@ -6,6 +6,13 @@
     upgrade: safe
 - name: Upgrade all installed snaps  # noqa no-changed-when
   command: snap refresh
+- name: Purge old snaps
+  shell: |
+    set -o pipefail && snap list --all | awk '/disabled/{print $1, $3}' |
+    while read snapname revision; do
+      snap remove "$snapname" --revision="$revision"
+    done
+  changed_when: false
```
kylelaker, line 15:
```suggestion
```

I'd almost rather assume that this always changes something

roles/oem/tasks/ubuntu_only.yml
```diff
@@ -6,6 +6,13 @@
     upgrade: safe
 - name: Upgrade all installed snaps  # noqa no-changed-when
   command: snap refresh
+- name: Purge old snaps
+  shell: |
+    set -o pipefail && snap list --all | awk '/disabled/{print $1, $3}' |
+    while read snapname revision; do
+      snap remove "$snapname" --revision="$revision"
+    done
```
kylelaker, line 14: I believe that this will only end up reporting as a failure if the last command in the `while` loop exits non-zero. For example, see:

```bash
set -o pipefail && cat << EOF | while read foo bar; do ( echo $foo $bar; exit $bar ); done
a 1
b 0
c 0
d 0
EOF
```

and the only scenario where this will set `$?` to `0` is if the last line has a non-zero value for the second item. In addition to setting `pipefail` it may be possible to also set `errexit`/`-e`; however, that would result in subsequent revisions not being deleted... The structure also results in additional sub shells due to the pipes.

I'm not 100% sure but perhaps something like:
```suggestion
    errors=0
    while read snapname revision; do
      snap remove "$snapname" --revision="$revision"
      errors=$((errors + $?))
    done <<< "$(snap list --all | awk '/disabled/{print $1, $3}')"
    exit $errors
```

Adding the error counter to the existing while loop (without also switching to a heredoc) resulted in failing to actually count the errors due to the sub shell. I used the following to validate the new suggestion since idk how to actually simulate a `snap remove` failure:

```bash
function create_values() {
cat << EOF
a 1
b 0
c 1
d 0
EOF
}

errors=0
set -o pipefail && while read foo bar; do
    ( echo $foo $bar; exit $bar )
    errors=$((errors + $?))
done <<< "$(create_values)"
exit $errors
```

I also didn't test this on a Ubuntu-ish directly so feel free to tell me if my examples are invalid in different versions of bash (or also if I am wrong generally).

kylelaker (changes requested):
Some thoughts/questions on making sure that we actually report failures correctly.

roles/oem/tasks/ubuntu_only.yml
ripleymj: Ansible lint still yells if you don't set pipefail, so why not. Also realized while we're exploding the loop, we can just track changes and report based on that instead of forcing it off or on.

kylelaker:
With the merge of #466 this now has a conflict but I think I am good with moving forward on merging.

ripleymj:
Should be good now, and cleaned a bit of the git history

kylelaker (approved):

PR #461, opened 2021-06-15
Update Linux Mint beta testing vars

ripleymj:
Untested as of yet, just getting a placeholder out here


ripleymj:
Builds, boots, and runs common role. I think that's enough for a beta to begin.

kylelaker (approved):

Issue #462, opened 2021-06-15
Uma beta reports our "apt configuration is corrupt"


ripleymj:
Uma beta reports our "apt configuration is corrupt" via tray icon. No further details are given yet. Possibly not even our problem to solve.

![image](https://user-images.githubusercontent.com/12940106/122121688-ece65180-cdf9-11eb-8fa0-a12f78bfc2f2.png)

ripleymj:
On 2nd glance, mintupdate appears to die with the following error

```
-- Exception occurred in the refresh thread:
Traceback (most recent call last):
  File "/usr/lib/linuxmint/mintUpdate/mintUpdate.py", line 830, in run
    self.application.builder.get_object("label_error_details").set_text(error_msg)
TypeError: Argument 1 does not allow None as a value
```

but the problem resolves itself after the first `apt-get update` is run manually.

ripleymj:
Reported as https://github.com/linuxmint/mint20.2-beta/issues/7

ripleymj:
With mintupdate 5.7.7, the crash no longer happens, but mintupdate still gets itself confused with the missing /var/lib/apt/lists/lock file. With the lock file in place but the lists missing, it still reports an error, but will eventually resolves itself once it resyncs package lists.

PR #464, opened 2021-06-23
Preserve apt lock file so mintupdate doesn't completely lose its mind

ripleymj:
Partial resolution for #462, but I still need to test this and hope I find a better solution.


ripleymj:
Lets get this in so we can get #466 in, and we can always patch it again later if a better solution is found

kylelaker (approved):
I'd like to see if there's a better way to make Uma happy but in the spirit of not letting perfect be the enemy of good and making sure we've got Uma support, I'm merging this.

PR #466, opened 2021-07-10
Update to Mint 20.2 Uma and remove outdated backgrounds

ripleymj:
We need to decide if #464 is good enough before merging this.


kylelaker (approved):
Oops guess we forgot to do the background change for Ulyssa :) Are there any other packages that we should consider removing? The IPTV stuff? Warpinator?

Approving and merging because I don't want to hold this up while having that discussion.

PR #471, opened 2021-08-27
Fix pylint errors

kylelaker:
This resolves some issues with the linter that were noticed after merging #470 which didn't have status checks run for some reason


kylelaker:
I've only been able to test on Ubuntu 20.04 on ARM64

ripleymj (approved):
Makes the linter happy and seems to run fine on Mint x64

PR #472, opened 2021-08-28
Hold Mesa packages to prevent white-out bug

ripleymj:
We've missed a serious bug so far, and I kind of hate this patch, but we need to get something published soon. Ubuntu recently updated the mesa drivers and they don't work when VirtualBox 3d acceleration is enabled. 

- https://forums.linuxmint.com/viewtopic.php?t=354405
- https://gitlab.freedesktop.org/mesa/mesa/-/issues/5192
- https://github.com/linuxmint/linuxmint/issues/409

Pinning the packages allows us to ship a working image for now, and we can backport an "un-hold" when the problem is resolved.

### DIFF
--- a/roles/oem/tasks/vm_only_pre.yml
+++ b/roles/oem/tasks/vm_only_pre.yml
@@ -11,3 +11,8 @@
     path: /var/lib/dpkg/arch
     state: absent
     line: 'i386'
+- name: Hold Mesa packages
+  dpkg_selections:
+    name: '{{ item }}'
+    selection: hold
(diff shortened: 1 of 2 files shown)


kylelaker (approved):
Reasonable to solution to an unfortunate problem. Our “unhold” is going to have to be somewhat carefully considered. We won’t just be able to delete this block (because for users who did get these held, they would remain as such) but I don’t want to eternally maintain a block that changes the `selection` for these specific packages. So perhaps the Uma branch just gets a different “backport” of the revert for this.

Issue #455, opened 2021-05-31
New `ansible-lint` warnings with `ignore_errors`


kylelaker:
Looks like `ansible-lint` is now grumpy about some unrelated things (but just grumpy; not failing) with `ignore_errors`. We can track this separately.

_Originally posted by @kylelaker in https://github.com/jmunixusers/cs-vm-build/issues/454#issuecomment-851543323_

Issue #446, opened 2021-02-20
Consider apt lock handling


ripleymj:
Ansible now has recommendations for handling an existing apt lock, which may be more elegant than our current methods: https://github.com/ansible/ansible/pull/73079

kylelaker:
I am not sure if this will work for us. It seems like the only place we "handle" a conflict over the lock file is in `ubuntu-build.json` where we try to do it prior to the `apt update` there. I am not convinced I want to start adding this to _every_ apt command throughout the repo either.

kylelaker:
It looks like the solution in that linked issue is no longer recommended and has been removed from the docs as of https://github.com/ansible/ansible/pull/76243. As of 2.12, instead specifying `lock_timeout` is preferred, though that already has a default of `60`. This is notably a bit different from outright stopping `unattended-upgrades.service` and infinitely waiting for `apt` to finish, but maybe it's enough for our use case?

PR #475, opened 2021-08-29
Resolve ignore-errors lint warnings

kylelaker:
This uses the apt lock workaround to replace an ignore_errors on update
and also just totally removes another ignore_errors for writing to the
.bashrcs of users. If that's failing, we probably need to make our
filter a bit better.

Closes: #455
Closes: #446


kylelaker:
This is in draft until I have a chance to test it.

kylelaker:
Validated on Ubuntu 20.04 ARM64

ripleymj:
I think the apt ignore failures was put in to handle when the CS mirror is down/unavailable. I tried putting mirror.cs.jmu.edu in /etc/hosts to simulate a failure, and Ansible refuses to proceed, even though the upstream mirror is a very viable choice.

kylelaker:
So to be clear, the apt cache is still being properly updated for the upstream Mint/Ubuntu repos? If `mirror.cs.jmu.edu` was down, would an install still successfully pull from the upstream repo? Basically, are we actually helping things along here or are we delaying a failure?

ripleymj:
Confirmed, an installation will complete as long as there's a viable mirror. I'd prefer we maintain this behavior.

kylelaker:
`mirror://` mostly works except for a message in `mintupdate` that the mirror is unreachable. https://github.com/linuxmint/mintupdate/pull/701 exists to try to help with that but we'll likely either need another solution or to just hold off until we can confirm better compatibility with `mintupdate`.

kylelaker:
Alright and I think this one is actually ready for review too. Tested by just blocking 134.126.0.0/16. It took a second to timeout but it eventually figured itself out, ignored the "down" mirror, and proceeded. No complaints from mintupdate on config format or anything like that.

PR #482, opened 2021-09-18
Spring 2022 Updates for wrapper script

kylelaker:
This performs some fairly significant refactoring to the wrapper script. It's incomplete but I am opening it as a draft early to collect additional ideas or objections.

- [X] Experimental course support
- [X] Add Settings for all config values
- [X] Properly follow XDG Base Directory standard
- [X] Use YAML instead of JSON
- [x] ~Improve error message logic~
- [x] ~Improve UX of VTE~
- [x] Link to VM readme in Help menu
- [x] Investigate `sudo` + `AskPass` helper instead of `pkexec`
  - Would make it easier to target which steps run with `become`

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -68,7 +68,7 @@ jobs:
           sudo apt-get -q install python3-gi-cairo gir1.2-gtk-3.0 libgirepository1.0-dev gir1.2-vte-2.91
       - name: Install dependencies
         run: |
-          pip3 install pyyaml pygobject pylint click PyGithub jinja2 aiohttp
+          pip3 install pyyaml pygobject pylint click PyGithub jinja2 aiohttp pyxdg
       - name: Run pylint
         run: |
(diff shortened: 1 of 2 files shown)


kylelaker:
`sudo` + `AskPass` doesn't really seem to be viable here

kylelaker:
Also probably want to kick error logic and VTE stuff to another semester.

kylelaker:
Going to move this back to draft until we see what's up with GTK 3/4 on Ubuntu 22.04/Mint 21.

kylelaker:
Alright this should be good to go. Seems to work with the default packages in Mint and Ubuntu (unless I've done something to invalidate that assumption).

Big features are:
 - Support for experimental courses (added when we talked about needing testers for Robotics)
 - Using XDG directories correctly for configuration and now that pyYAML is available my default we can use YAML for the config
 - Makes everything we had a config value for available in settings (like allowing using the `main` branch)

ripleymj (approved):

PR #485, opened 2021-12-11
Una beta

ripleymj:
Update URL for Mint Una beta, and remove VirtualBox Mesa hack since that seems to resolved now.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -33,7 +33,7 @@ jobs:
           python-version: 3.x
       - name: Install dependencies
         run: |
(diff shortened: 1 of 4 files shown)


ripleymj:
I'll work on the Packer failures later

ripleymj:
I think the Packer issues were a bug in a nightly build, and have resolved themselves. However, an update to the Actions image, has broken demjson against Python3, so I've bumped that to a 3-compatible version.

Now it seems eclipse.org has taken the morning off.

kylelaker (approved):

PR #489, opened 2022-03-26
Update Eclipse to 2022-03

kylelaker:
This version was released on March 16. Hopefully the mirrors have had
time to catch up.


kylelaker:
🤷🏻 😞 

Well... at least the Eclipse mirrors have figured themselves out! Looks like we'll have to poke at the Packer nightly build URL and some of the Ubuntu package dependencies... Also `ansible-lint` wants us to use the fully qualified names for modules now -- does that even work with Mint's version of Ansible?

ripleymj (approved):

PR #490, opened 2022-03-26
fix lints

kylelaker:
- Update package cache before apt installs
- Use newest versions of workflow actions
- Add Dependabot for GitHub Actions


kylelaker:
Going to open as a PR against @ripleymj's other fixes in #490

PR #491, opened 2022-03-26
Fix lint

ripleymj:
I too had fallen into a trap of fixing lints. I'm not confident enough to propose merging this yet, but would probably like to get #490 rebased under it.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python 3.x
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.x
       - name: Install dependencies
@@ -28,7 +28,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Set up Python 3.x
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.x
       - name: Install dependencies
@@ -37,15 +37,23 @@ jobs:
       - name: Run jsonlint
         run: |
           jsonlint packer/*json
-      - name: Fetch and unzip packer nightly
+      - name: Install Hashicorp repo
+        run: >
+          wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor |
+          sudo tee /usr/share/keyrings/hashicorp-archive-keyring.gpg &&
+          echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg]
+          https://apt.releases.hashicorp.com $(lsb_release -cs) main" |
+          sudo tee /etc/apt/sources.list.d/hashicorp.list
+      - name: Install Packer
         run: |
-          wget -nv -O $HOME/packer.zip https://github.com/hashicorp/packer/releases/download/nightly/packer_linux_amd64.zip
-          unzip -d $HOME/packer $HOME/packer.zip
+          sudo apt-get update
+          sudo apt-get install -y -qq packer
       - name: Packer validate
         run: |
-          $HOME/packer/packer validate packer/ubuntu-build.json
-          $HOME/packer/packer validate packer/mint-build.json
-          $HOME/packer/packer validate -var-file=packer/beta-vars.json packer/mint-build.json
+          packer validate packer/ubuntu-build.json
+          packer validate packer/mint-build.json
+          # Allow beta/test builds to fail validation
+          packer validate -var-file=packer/beta-vars.json packer/mint-build.json || true
   Python:
     name: Run Python lint tests
     runs-on: ubuntu-20.04
@@ -53,11 +61,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Set up Python 3.x
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.x
       - name: Install Ubuntu dependencies
         run: |
+          sudo apt-get update
           sudo apt-get -q install python3-gi-cairo gir1.2-gtk-3.0 libgirepository1.0-dev gir1.2-vte-2.91
       - name: Install dependencies
         run: |
@@ -89,7 +98,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Set up Python 3.x
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.x
       - name: Install dependencies
@@ -111,7 +120,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Set up Python 3.x
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.x
       - name: Install dependencies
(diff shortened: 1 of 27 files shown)


packer/ubuntu-build.json
```diff
@@ -11,7 +11,7 @@
     "audio": "pulse",
     "output_dir": "{{pwd}}/artifacts_ubuntu",
     "mirror_url": "https://mirror.cs.jmu.edu/pub/ubuntu-iso/focal",
-    "iso_file": "ubuntu-20.04.3-desktop-amd64.iso",
+    "iso_file": "ubuntu-20.04.4-desktop-amd64.iso",
```
kylelaker, line 14: Was wondering why our automated workflow didn't catch this...

image

😢

roles/oem/tasks/vm_only_post.yml
```diff
@@ -53,11 +53,11 @@
 # Extensions which are non-free software and available only under the Oracle
 # PUEL. We should ensure we have not accidentally included them.
 - name: Validate missing USB 2/3 controllers  # noqa risky-shell-pipe
-  command: lspci
+  ansible.builtin.command: lspci
```
kylelaker, line 56: Looks like a few other uses have been refactored to use the longer form. Should we be consistent about that? I'll mark the others as suggestions in case that's the direction we want to go.

```suggestion
  ansible.builtin.command:
    cmd: lspci
```

roles/oem/tasks/vm_only_post.yml
```diff
@@ -30,21 +30,21 @@
 # The apt database can be fairly large and make the image unnecessarily large
 # so removing it before shipping allows us to ship a smaller image.
 - name: Clean the apt package cache
-  command: apt-get clean
+  ansible.builtin.command: apt-get clean
```
kylelaker, line 33:
```suggestion
  ansible.builtin.command:
    cmd: apt-get clean
```

roles/oem/tasks/ubuntu_only.yml
```diff
@@ -2,12 +2,12 @@
 # tasks file for Ubuntu
 
 - name: Upgrade all installed packages
-  apt:
+  ansible.builtin.apt:
     upgrade: safe
 - name: Upgrade all installed snaps  # noqa no-changed-when
-  command: snap refresh
+  ansible.builtin.command: snap refresh
```
kylelaker, line 8:
```suggestion
  ansible.builtin.command:
    cmd: snap refresh
```

roles/oem/tasks/mint_only.yml
```diff
@@ -2,35 +2,35 @@
 # tasks file for Linux Mint
 
 - name: Update mintupdate package
-  apt:
+  ansible.builtin.apt:
     name: "{{ upgrade_pre_mintupdate }}"
     state: latest
 - name: Upgrade all installed packages  # noqa no-changed-when
-  command: /usr/bin/mintupdate-cli -y upgrade
+  ansible.builtin.command: /usr/bin/mintupdate-cli -y upgrade
```
kylelaker, line 9:
```suggestion
  ansible.builtin.command:
    cmd: /usr/bin/mintupdate-cli -y upgrade
```

roles/oem/handlers/main.yml
```diff
@@ -1,4 +1,4 @@
 ---
 # handlers file for oem
 - name: Update dconf database
-  command: /usr/bin/dconf update
+  ansible.builtin.command: /usr/bin/dconf update
```
kylelaker, line 4:
```suggestion
  ansible.builtin.command:
    cmd: /usr/bin/dconf update
```

roles/oem/tasks/vm_only_post.yml
```diff
@@ -53,11 +53,11 @@
...
 - name: Validate missing USB 2/3 controllers  # noqa risky-shell-pipe
-  command: lspci
+  ansible.builtin.command: lspci
   register: lspci_output
   failed_when: "'ECHI' in lspci_output.stdout or 'xHCI' in lspci_output.stdout"
   changed_when: false
 
 - name: Fill disk with zeros to assist with compression
-  shell: dd if=/dev/zero of=/bigfile bs=1M; sync; rm /bigfile
+  ansible.builtin.shell: dd if=/dev/zero of=/bigfile bs=1M; sync; rm /bigfile
```
kylelaker, line 62:
```suggestion
  ansible.builtin.shell:
    cmd: dd if=/dev/zero of=/bigfile bs=1M; sync; rm /bigfile
```

roles/user/handlers/main.yml
```diff
@@ -1,10 +1,10 @@
 ---
 # handlers file for user
 - name: Update desktop menu
-  command:
+  ansible.builtin.command:
     /usr/bin/update-desktop-database
```
kylelaker, line 5:
```suggestion
  ansible.builtin.command:
    cmd: /usr/bin/update-desktop-database
```

roles/user/handlers/main.yml
```diff
@@ -1,10 +1,10 @@
...
 - name: Suggest restart
-  command: >-
+  ansible.builtin.command: >-
     notify-send -u critical "JMU Software Change"
     "Changes have been made to your machine that
     will only take effect when you log out or
```
kylelaker, line 10:
```suggestion
  ansible.builtin.command:
    cmd: >-
      notify-send -u critical "JMU Software Change"
      "Changes have been made to your machine that
      will only take effect when you log out or
```

And GitHub won't let me comment it but the next line would need to be indented to...

.github/workflows/lint.yml
```diff
@@ -37,27 +37,35 @@ jobs:
...
-        run: |
-          wget -nv -O $HOME/packer.zip https://github.com/hashicorp/packer/releases/download/nightly/packer_linux_amd64.zip
-          unzip -d $HOME/packer $HOME/packer.zip
+      - name: Install Hashicorp repo
+        run: >
+          wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor |
+          sudo tee /usr/share/keyrings/hashicorp-archive-keyring.gpg &&
+          echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg]
+          https://apt.releases.hashicorp.com $(lsb_release -cs) main" |
+          sudo tee /etc/apt/sources.list.d/hashicorp.list
```
kylelaker, line 46: Since we're already running as root, we can trim this and remove some of the calls to `tee` to more closely match the commands at https://wiki.debian.org/DebianRepository/UseThirdParty. Honestly... that page mentions that de-armoring is only required pre-`stretch`, so we may be able to just `wget` directly to `$KEYRING_FILE` but ehh.

```suggestion
        run: |
          KEYRING_FILE="/usr/share/keyrings/hashicorp-archive-keyring.gpg"
          curl -SL https://apt.releases.hashicorp.com/gpg | gpg --dearmor > "$KEYRING_FILE"
          echo "deb [signed-by=$KEYRING_FILE] https://apt.releases.hashicorp.com $(lsb_release -cs) main" > /etc/apt/sources.list.d/hashicorp.list
```

ripleymj: I'm not convinced we're actually root by default. This failed hard when I tried it.

ripleymj:
I think this is about done for now.

- shell/command fails at this point, I think due to https://github.com/ansible/ansible/pull/71824. Because I'd already edited them, then put them back, I left it as two commits for an easy revert later.
- Until that's resolved, I demoted the linter error to warning. Ansible-lint also recommends not erroring on experimental rules, which sounds reasonable to me
- dconf is in community.general, not ansible.base, but that's packaged in 20.04 yet
- ansible-lint now does yamllint too but skips GitHub Actions files, so I merged those Actions

kylelaker (approved):

PR #493, opened 2022-04-03
Add SPICE tools

kylelaker:
This pulls a commit from #457 and also adds `spice-webdavd` per https://mac.getutm.app/support/ to support folder sharing.

I want to play around a bit with packaging for `libvirt` and other tools and having this shared code already in `main` would probably help both branches.

I was tempted to remove the `when` on this, to just always install in a VM since the SPICE tools aren't all that heavy and it shouldn't hurt the VBox builds to have them installed.

### DIFF
--- a/roles/oem/tasks/vm_only_post.yml
+++ b/roles/oem/tasks/vm_only_post.yml
@@ -13,6 +13,12 @@
   notify:
     - Update dconf database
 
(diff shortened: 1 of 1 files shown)


ripleymj (approved):
I'm not deeply opposed to removing the when, depending on where your libvirt adventures take you

PR #497, opened 2022-06-26
Fix var-spacing lint error

ripleymj:
Yet another gift from ansible-lint, found in updating #457


kylelaker (approved):

kylelaker:
Hmm maybe someday we can pin the linter version. But also, it feels like that just delays the maintenance or makes it more likely to forget... Thoughts for another day though.

Thanks for this!

PR #501, opened 2022-07-17
Convert Packer definition to HCL2

### LINKED ISSUE #460: Move configurations out of vboxmanage
Packer 1.7 introduces additional VirtualBox options so we can remove a few manual `vboxmanage` commands: https://github.com/hashicorp/packer/pull/10671

kylelaker:
This rewrites the packer JSON files to use the HCL2 language support
added in 1.5 and stabilized in 1.7. The initial conversion was performed
using the `hcl2_upgrade` command but quite a lot of custom work has been
done on top of that.

A base shared definition for the VM is created that defines all the
shared options (CPU, memory, SSH stuff, etc) and then that is
specialized in the `build` block for each `source`, this is where the
differences between Ubuntu and Mint are specified.

Additionally, we now store version information as complex objects that
meet our needs a bit more closely. This is required because some of the
interpolation that we did in JSON is no longer available.

As I was reviewing the VirtualBox builder config I also moved a few
things from manual `VBoxManage` commands to actual packer definitions
but some of those could have been done in the JSON format as well.

Closes #460


kylelaker:
Hmm... Looks like something goes wrong either during the Ubuntu `boot_command` or with the preseed file. Not sure if that same behavior happens on `main` or not atm so I'll have to check.

kylelaker:
@ripleymj If you're opposed to enabling EFI, I can try to debug the second commit on this PR a bit more... but also might be fun to just see what happens?

ripleymj:
It looks like the Ubuntu file never got updated for 22.04, and the corresponding grub syntax change. I think you've got the right syntax now, especially `c<wait>` with or without EFI.

kylelaker:
Okay. Let me try just dropping the EFI change then. For some reason it felt like I got a slightly different flow with BIOS but let's find out!

kylelaker:
> got the right syntax now

u right. EFI removed (for now)

.github/workflows/lint.yml
```diff
@@ -37,9 +37,6 @@ jobs:
       - name: Install dependencies
         run: |
           pip3 install demjson3
```
kylelaker, line 39: Do we need this (and python itself) for any other reason? Or should I remove those?

kylelaker (commented):
Okay I've done my best here to also reduce some code duplication; it makes less-than-perfectly clean code in a few places but:
 - we just try both `initrd` and `initrd.lz` as paths for the initrd; grub ignores the wrong one
 - we use `|| true` to a systemctl command so that we can use the same script to stop `apt` and run the `oem-build` script

The `mint_info` and `ubuntu_info` blocks in `locals` also aren't the mostly lovely thing in the world. But on the whole, this significantly reduces the duplication that we had before and makes these a little easier to manage going forward hopefully.

I wish there was a good way to `merge` the `export_opts` so that we only had to specify the different values but I don't see a phenomenal option (and again, we were duplicating it before anyway).

Unless I find some sort of issue with this during testing, I think that for the most part I am done making changes here and this is _actually_ ready for review.

.github/workflows/lint.yml
ripleymj: I think it supplies jsonlint, so I guess if you're killing things, be thorough.

ripleymj:
Before I try to understand this, any thoughts on how AARCH64 from #457 will merge in? Is qemu too far away to merge with VirtualBox?

kylelaker:
> Before I try to understand this, any thoughts on how AARCH64 from #457 will merge in? Is qemu too far away to merge with VirtualBox?

Yeah, so it'd definitely need a different top-level `source` block since it uses a different builder plugin. We _might_ be able to reuse the `source` block? And there'd be some overlap... I think we'd want to look at https://github.com/hashicorp/packer/blob/03d29386dd92d45ad3ada54dc282f229f7bed526/examples/hcl/ubuntu/build.pkr.hcl as an example (and it's sibling files in that directory).

.github/workflows/lint.yml
```diff
@@ -53,10 +43,10 @@ jobs:
           sudo apt-get install -y -qq packer
       - name: Packer validate
         run: |
-          packer validate packer/ubuntu-build.json
-          packer validate packer/mint-build.json
+          packer validate .
           # Allow beta/test builds to fail validation
-          packer validate -var-file=packer/beta-vars.json packer/mint-build.json || true
+          packer validate -var-file=packer/mint-beta.pkrvars.hcl . || true
+        working-directory: packer
```
ripleymj, line 49: How would you feel about adding a step like `packer fmt -diff -check packer/` to check formatting now that jsonlint is gone?

kylelaker: Done. I used `working-directory` again so that it looks the same in both steps. And I reordered the commands because I don't remember if GH steps are effectively `set -e` and I don't think we should need to be able to remember :)

ripleymj (approved):

.github/workflows/lint.yml
```diff
@@ -53,10 +43,15 @@ jobs:
           sudo apt-get install -y -qq packer
       - name: Packer validate
         run: |
-          packer validate packer/ubuntu-build.json
-          packer validate packer/mint-build.json
           # Allow beta/test builds to fail validation
-          packer validate -var-file=packer/beta-vars.json packer/mint-build.json || true
+          packer validate -var-file=packer/mint-beta.pkrvars.hcl . || true
```
ripleymj, line 47: I just noticed this step fails because we're already in the packer directory, so the file reference is invalid.

ripleymj (changes requested):

.github/workflows/lint.yml
kylelaker: Ahh damn `|| true`!

ripleymj (approved):

PR #504, opened 2022-07-19
Realign courses

### LINKED ISSUE #503: Update course configurations for Fall 2022
Some courses have had some pretty significant changes (especially CS 149). These updates will need to be incorporated into the VM before we release this fall.

I think this work is already in progress but I wanted to create an issue to track/discuss it.

ripleymj:
Based on the 149 conversation I've seen, and walking through the 101 lab exercises, this PR simplifies and consolidates the CS 101/149 content.

Closes #503


local.yml
```diff
@@ -7,11 +7,9 @@
...
+    - { role: basic_prog_pkgs, tags: ["cs101", "cs149"] }
     - { role: adv_prog_pkgs, tags: ["cs261", "cs361", "cs432"] }
-    - { role: eclipse, tags: ["cs101", "cs149", "cs159"] }
-    - { role: jgrasp, tags: ["cs149"] }
-    - { role: finch, tags: ["cs101"] }
+    - { role: eclipse, tags: ["cs159"] }
     - { role: y86, tags: ["cs261", "cs432"] }
     - { role: programming_langs, tags: ["cs430"] }
-    - { role: vscode, tags: ["cs149"] }
+    - { role: vscode, tags: ["cs159"] }
```
kylelaker, line 15: I am starting to wonder if we shouldn't just always make VS Code available... you're closer to the students/department than I am. But it feels like a pretty "standard" IDE regardless of which course someone may be in

roles/basic_prog_pkgs/vars/main.yml
```diff
@@ -1,18 +1,13 @@
...
  - logisim
  - meld
+ - sqlite
  - sqlitebrowser
  - thonny
- - wireshark
+
+basic_prog_pkgs_pip:
+ - flake8-docstrings
+ - darglint
```
kylelaker, line 13: At least the former is available as `python3-flake8-docstrings` in Ubuntu. Not sure about `darglint`. But it might be nice to use the system Python in this course?

roles/basic_prog_pkgs/tasks/main.yml
```diff
@@ -5,27 +5,10 @@
...
-  loop: "{{ real_users }}"
-- name: Set Idle pref in existing mimeapps
-  ansible.builtin.lineinfile:
-    path: '{{ item.homedir }}/.config/mimeapps.list'
-    regexp: '^text/x-python='
-    line: 'text/x-python=idle-python3.6.desktop;gedit.desktop;pluma.desktop'
+- name: Install intro pip packages
+  ansible.builtin.pip:
+    name: '{{ basic_prog_pkgs_pip }}'
+    state: present
```
kylelaker, line 11:
```suggestion
  ansible.builtin.pip:
    name: '{{ basic_prog_pkgs_pip }}'
    state: present
    extra_args: --user
```

scripts/hashlint.py
```diff
@@ -23,14 +23,6 @@
...
-        "hash": "jgrasp.hash",
-        "urls": ["jgrasp.url"],
-    },
-    "roles/finch/vars/main.yml": {
-        "hash": "finch.hash",
-        "urls": ["finch.url"],
-    },
     "roles/eclipse/vars/main.yml": {
         "hash": "eclipse.hash",
         "urls": ["eclipse.url", "eclipse.url_backup"],
```
kylelaker, line 28: Well this script feels over-engineered now.

roles/basic_prog_pkgs/tasks/main.yml
kylelaker: Do we set `PATH` stuff correctly to handle passing `--user`?

kylelaker (commented):
Not sure how far you are in this but I've got a few initial thoughts. We could merge as-is but I think we have some time to discuss these.

roles/basic_prog_pkgs/tasks/main.yml
ripleymj: Pip appears to be handling this correctly for a user install, perhaps since https://github.com/pypa/pip/pull/7002. Not sure I understand the PATH reference, but the user install is writing to .local/bin, which does not appear to be in PATH

roles/basic_prog_pkgs/vars/main.yml
ripleymj: I don't see darglint, but Thonny does appear to recognize system packages in its "Package Manager" so I can split this.

roles/basic_prog_pkgs/tasks/main.yml
kylelaker: Hmm yeah, not having `~/.local/bin` in `$PATH` is probably fine for now with the tool(s) we've got at the moment. But it's probably something to be aware of in the future.

roles/basic_prog_pkgs/vars/main.yml
```diff
@@ -1,18 +1,14 @@
...
- - fonts-crosextra-carlito
- - geany
- - idle
+ - flake8
  - libreoffice
- - libreoffice-templates
- - libreoffice-style-tango
  - logisim
  - meld
+ - python3-flake8-docstrings
```
kylelaker, line 8:
```suggestion
 - python3-flake8-docstrings
 - python-is-python3
```

Maybe it's not ideal from a like correctness and portability perspective but it _does_ seem like it'd be more user friendly? Maybe we should even add this to `common` or `user`?

kylelaker (commented):
Overall I think this is basically perfect but for usability and friendliness to new users I wonder if we should add:

 - `ipython3`
 - `python-is-python3`

kylelaker (dismissed):
Actually, I don't think my Python thoughts are worth holding up review over. Going to approve but just not click the "merge" button. Feel free to tell me no and to merge or to add the packages. Because either path feels equally valid.

roles/basic_prog_pkgs/vars/main.yml
ripleymj: I don't want to get into installing tools outside a course's recommendations, but these seemed like helpful cleanups of what we're offering.

kylelaker (approved):

PR #508, opened 2022-08-11
Update to 22.04.1 ISO

ripleymj:
Me filling in as the auto-updater until we find a way to repair it.

### DIFF
--- a/packer/variables.pkr.hcl
+++ b/packer/variables.pkr.hcl
@@ -15,7 +15,7 @@ variable "ubuntu_version" {
     minor_version = string
   })
   default = {
-    version       = "22.04"
+    version       = "22.04.1"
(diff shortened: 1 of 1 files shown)


packer/variables.pkr.hcl
```diff
@@ -15,7 +15,7 @@ variable "ubuntu_version" {
     minor_version = string
   })
   default = {
-    version       = "22.04"
+    version       = "22.04.1"
     minor_version = ""
```
kylelaker, line 19: Hmm I had sorta expected we'd use the `minor_version` for this. But honestly I didn't realize that this would work too!

kylelaker:
Can we try #509 and come back to this if that still doesn't work out?

PR #509, opened 2022-08-12
Fix the Ubuntu update workflow to work with HCL

kylelaker:
The change to HCL broke the update workflow because it's not JSON and
the JSON file we were parsing was deleted. This uses an
`auto.pkrvars.hcl` file to set the Ubuntu version (for which an updated
version is intentionally not committed so that the workflow will have
work to do). This means that check will fail until the workflow runs and
opens its own PR (becuse validation fails if the checksum can't be
found).

This is done _in addition_ to `version` so that the the friendlier
version can be used in contexts where the patch version doesn't really
matter (like the parent directory or the build metadata).

### DIFF
--- a/.github/workflows/auto-update-iso.yml
+++ b/.github/workflows/auto-update-iso.yml
@@ -5,7 +5,7 @@ on:
     - cron: '0 9 * * *'
   workflow_dispatch:
 env:
-  UBUNTU_RELEASE: focal
+  UBUNTU_RELEASE: jammy
 jobs:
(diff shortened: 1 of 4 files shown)


ripleymj (approved):

ripleymj:
Merge when you're ready, and I assume we just see what happens in the AM?

kylelaker:
> I assume we just see what happens in the AM

I was going to manually re-enable and execute with the `workflow_dispatch` trigger.

PR #515, opened 2022-10-16
Use new GitHub Actions output mechanism

kylelaker:
GitHub has [recently deprececated][deprecation notice] the `set-output`
function in workflows, instead having users switch to to using a newly
created environment file for this purpose. Because this is confusing,
the behavior is extracted to a function that we can reuse.

[deprecation notice]: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/scripts/update-ubuntu-release
+++ b/scripts/update-ubuntu-release
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+set-output() (
+  local name="$1"
+  local value="$2"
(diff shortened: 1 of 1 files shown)


kylelaker:
This new `ansible-lint` error doesn't feel fun:

```
Use `ansible.builtin.include_tasks` or `ansible.legacy.include_tasks` instead.
```

ripleymj (approved):
Sorry, I missed that you rebased this

PR #517, opened 2022-12-16
Toggle VBox7 NAT filter to allow preseed file fetching

ripleymj:
From https://github.com/hashicorp/packer/issues/12118, this seems to be what is jamming us up on VBox7 builds. Went ahead and added a post-build step to restore the default since I don't think we have any other use for it being enabled.


kylelaker (approved):
Nice find! Unfortunate to need the workaround but glad it solves it!

PR #520, opened 2022-12-28
Vera update

ripleymj:
Update the build to Vera 21.1. There is some bug where oem-config isn't being installed, so I've manually patched it in for now. I'd prefer to research further before merging, but this allows anyone to experiment at the moment


kylelaker (dismissed):
Not sure how I missed this! Thanks for tracking this!

kylelaker (approved):

kylelaker:
Thanks for this & getting it updated!

PR #528, opened 2023-06-25
Allow EHCI and xHCI controllers now that they're open source

ripleymj:
Still testing this one, but since the USB controllers were open-sourced with the VBox 7 release a year ago, I think it's safe to leave them to their defaults instead of forcing them off.


ripleymj:
This builds and imports correctly, and appears to still prefer the USB 1.1 controller.

PR #529, opened 2023-07-17
Victoria release

ripleymj:
(closed beta PR, in favor of updating all at once because of the shared ansible fix)

Update beta and release variables for Victoria release

Also, revert apt-get clean step that breaks the build, because some fool (https://github.com/jmunixusers/cs-vm-build/pull/523) tried to fix a linter error that requires Ansible 2.13 while we're stuck with 2.10.


kylelaker (approved):
I'd blame that fool if Ansible version numbers made any sense to me these days :) 

Thanks for getting this ready to go!

PR #532, opened 2023-10-04
Ansible lint vars

ripleymj:
ansible-lint has decided vars need to be prefixed with the role name (https://github.com/ripleymj/cs-vm-build/actions/runs/6373527962/job/17297317317) so this tries to clean that up. Unfortunately, eclipse looks kind of goofy when you name everything "eclipse_eclipse", so I squashed that one for now but am open to suggestions


ripleymj:
Changed my mind, going to merge this as is. I need to do the Eclipse 2023-09 release info as well so I'll mess that whole role up at once.

PR #535, opened 2023-12-31
Fix packer validate plugin error

kylelaker:
It seems like Packer doesn't realize that we need the Virtualbox plugin
when it does the validation.


kylelaker:
@ripleymj I think I need you to delete my source branch here... 😳

PR #538, opened 2024-01-13
Update to Mint 21.3 Virginia

kylelaker:
This updates to Mint 21.3. As part of the release, Mint changed to using
syslinux on BIOS installs. I really don't want to learn how that works
for the boot command so instead, this migrates to UEFI which Mint
promises is better supported in 21.3 and which VirtualBox continually is
improving support for. The boot command gets some minor tweaks but these
match what was planned for QEMU on AARCH64 anyway.

Supersedes #419

### DIFF
--- a/packer/http/oem-preseed.cfg
+++ b/packer/http/oem-preseed.cfg
@@ -36,13 +36,6 @@ d-i partman/choose_partition select finish
 d-i partman/confirm boolean true
 d-i partman/confirm_nooverwrite boolean true
 
-# This is fairly safe to set, it makes grub install automatically to the MBR
-# if no other operating system is detected on the machine.
-d-i grub-installer/only_debian boolean true
-
-# To install to the first device (assuming it is not a USB stick):
-d-i grub-installer/bootdev  string default
-
 ### Clock and time zone setup
 d-i clock-setup/utc boolean true
 d-i time/zone string US/Eastern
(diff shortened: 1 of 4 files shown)


packer/http/oem-preseed.cfg
```diff
@@ -36,13 +36,6 @@ d-i partman/choose_partition select finish
 d-i partman/confirm boolean true
 d-i partman/confirm_nooverwrite boolean true
 
-# This is fairly safe to set, it makes grub install automatically to the MBR
-# if no other operating system is detected on the machine.
-d-i grub-installer/only_debian boolean true
-
-# To install to the first device (assuming it is not a USB stick):
-d-i grub-installer/bootdev  string default
-
```
kylelaker, line 45: These seem unnecessary with the switch to EFI

packer/main.pkr.hcl
```diff
@@ -33,17 +34,23 @@ source "virtualbox-iso" "base-build" {
...
   boot_wait = "5s"
   boot_command = [
-    "<esc><wait><esc><wait><esc><wait>",
-    "c<wait>",
-    "linux /casper/vmlinuz fsck.mode=skip<wait> noprompt",
+    # Enter the command line
+    "c<wait><wait>",
+    # Configure the kernel
+    "linux /casper/vmlinuz",
+    " boot=casper",
```
kylelaker, line 41: Without this new flag I ended up at an initramfs prompt

ripleymj (approved):

PR #541, opened 2024-01-13
Enable good USB

kylelaker:
We previously started to allow XHCI with them being included in the base VirtualBox and no longer restrictively licensed; let's go ahead and actually turn them on all the way after we're done with the initial build. Even though it's unlikely too many things will get mapped into the VM, it's nice to have a sensible and good default.

I tested this by running a 4K webcam in Mint and it seemed to perform well.


ripleymj:
Can you hit this with a quick rebase when you get a moment?

ripleymj (approved):

PR #544, opened 2024-01-27
Use Mint 21.3 Edge image

kylelaker:
We probably don't need to start out with the 6.5 kernel but why should
that stop us from doing it anyway?

This changes the `beta` variable to a `build_type` variable (which
avoids needing to handle both edge/beta being set).


ripleymj (approved):
It looks like this may add couple hundred MB to final build, but we can tune that later if necessary

PR #546, opened 2024-02-04
Spring2024 cleanup

ripleymj:
I probably should've flagged the Sp24 string in #538, but here we are. Also, let's fix the wallpaper removal permanently.


roles/oem/tasks/vm_only_pre.yml
```diff
@@ -1,6 +1,19 @@
...
 # Tasks for the oem role to run at the beginning of playbook execution
+- name: Gather the package facts
+  ansible.builtin.package_facts:
+    manager: auto
+- name: Remove unneeded backgrounds
+  ansible.builtin.apt:
+    name: '{{ item }}'
+    state: absent
+    purge: yes
+  loop: "{{ ansible_facts.packages | select('search', 'mint-background') | list }}"
```
kylelaker, line 12: Can we pass this directly to `name` instead of needing to `loop`? iirc `ansible.builtin.apt` is supposed to be smart so maybe it doesn't matter either way.

kylelaker (approved):

ripleymj:
I like it in concept to avoid repeated apt calls, but it currently breaks the `ansible_distribution_release not in item` so I'll need to play more later

roles/oem/tasks/vm_only_pre.yml
```diff
@@ -1,6 +1,18 @@
...
+  ansible.builtin.package_facts:
+    manager: auto
+- name: Remove unneeded backgrounds
+  ansible.builtin.apt:
+    name: "{{ ansible_facts.packages | select('search', 'mint-background') | list }}"
+    state: absent
+    purge: yes
+  when:
+    - 'ansible_distribution_release not in item'
+    - 'ansible_distribution == "Linux Mint"'
```
kylelaker, line 14:
```suggestion
- name: Remove unneeded backgrounds
  ansible.builtin.apt:
    name: "{{ ansible_facts.packages | select('search', 'mint-background') | reject('search', ansible_distribution_release) | list }}"
    state: absent
    purge: yes
  when:
    - 'ansible_distribution == "Linux Mint"'
```

Not 100% sure on this (I haven't tested it) but maybe?

ripleymj: That's about where Google was leading me. Testing that now, but it also gives me a slightly different idea to try as well

ripleymj: Thoughts on the modified version now? One call to apt, and even though it's not guarded by `distro == mint` we never did that before. If you agree, I'll squash git before we merge

kylelaker: LGTM!

ripleymj: Done

kylelaker (approved):

PR #558, opened 2025-08-14
Update beta to 22.2 Zara

ripleymj:
Kernel panic on first boot after OEM, but other than that the basics seem to check out.


ripleymj:
Enough beta, on to the real version.

laurelmay (approved):

PR #564, opened 2026-05-05
Zena update

ripleymj:
Went in to bump the Mint version. Decided to bump Eclipse while I was there. Realized that with all roles installed, we only have 1.8GB left free, so bumped the disk size to 25GB. Re-ran the playbook and it died on VS Code because they install a conflicting modernized apt source file.

Anyways, things working as they used to and the world can go on ignoring this.

### DIFF
--- a/roles/vscode/tasks/main.yml
+++ b/roles/vscode/tasks/main.yml
@@ -4,17 +4,22 @@
   ansible.builtin.apt_key:
     id: BC528686B50D79E339D3721CEB3E94ADBE1229CF
     url: https://packages.microsoft.com/keys/microsoft.asc
-    keyring: /etc/apt/trusted.gpg.d/packages.microsoft.gpg
+    keyring: /usr/share/keyrings/microsoft.gpg
     state: present
 - name: Install VSCode Repository file
-  ansible.builtin.apt_repository:
-    repo: deb [arch=amd64,arm64,armhf signed-by=/etc/apt/trusted.gpg.d/packages.microsoft.gpg] https://packages.microsoft.com/repos/code stable main
-    state: present
-    filename: vscode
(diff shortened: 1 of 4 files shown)


roles/vscode/tasks/main.yml
```diff
@@ -4,17 +4,22 @@
...
+    uris: https://packages.microsoft.com/repos/code
+    suites: stable
+    components:
+      - main
+    signed_by: /usr/share/keyrings/microsoft.gpg
 - name: Install VSCode
   ansible.builtin.apt:
     name: 'code'
     state: latest
+    update_cache: yes
```
laurelmay, line 22:
```suggestion
  ansible.builtin.deb822_repository:
    name: vscode
    types: deb
    uris: https://packages.microsoft.com/repos/code
    suites: stable
    components:
      - main
    signed_by: /usr/share/keyrings/microsoft.gpg
  register: vscode_repository
- name: Install VSCode
  ansible.builtin.apt:
    name: 'code'
    state: latest
    update_cache: "{{ vscode_repository.changed }}"
```

Will `update_cache: yes` cause this to basically always report as changed? If so, can we conditionally do it when the repository step changes?

laurelmay (commented):
idk if this diff is gonna end up borked. github showed some of the craziest stuff i've ever seen

roles/vscode/tasks/main.yml
ripleymj: I didn't notice it reporting a change, but I'll retest before we merge

laurelmay (approved):

roles/vscode/tasks/main.yml
ripleymj: Doesn't report a change on a second run, seems the module behaves like you'd hope.

